PR #21, opened 2021-11-24
Support Array type

### LINKED ISSUE #20: Support Array type

pan3793:
close #20

### DIFF
--- a/clickhouse-spark-32-integration/src/test/scala/xenon/clickhouse/single/ClickHouseDataTypeSuite.scala
+++ b/clickhouse-spark-32-integration/src/test/scala/xenon/clickhouse/single/ClickHouseDataTypeSuite.scala
@@ -27,14 +27,17 @@ class ClickHouseDataTypeSuite extends BaseSparkSuite
     with Logging {
 
   test("write supported data types") {
-    val cols =
+    val structFields =
       StructField("id", DataTypes.LongType, false, Metadata.fromJson("""{"comment": "sort key"}""")) ::
         StructField("col_string", DataTypes.StringType, false) ::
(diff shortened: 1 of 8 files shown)


pan3793:
cc @flydreamwangping

clickhouse-spark-32-integration/src/test/scala/xenon/clickhouse/single/ClickHouseDataTypeSuite.scala
```diff
@@ -14,15 +14,46 @@
...
+  test("write supported data types") {
+    val structFields =
+      StructField("id", DataTypes.LongType, false, Metadata.fromJson("""{"comment": "sort key"}""")) ::
+        StructField("col_string", DataTypes.StringType, false) ::
+        StructField("col_array_string", DataTypes.createArrayType(DataTypes.StringType, false), false) :: Nil
+    val db = "t_w_s_db"
+    val tbl = "t_w_s_tbl"
+    withTable(db, tbl, structFields) {
+      val tblSchema = spark.table(s"$db.$tbl").schema
+      // TODO v2 create table should respect element nullable of array field
```
pan3793, line 38: Not sure if it is a spark bug or mine, further investigation is needed.

pan3793: It should not be considered as bug, but a limitation of SQL syntax.

For instance,
```sql
creata table xxx (abc  ARRAY<STRING> NOT NULL)
```
There is not way to mark the inner type `STRING` as `NULL` or `NOT NULL`, then it will be treat as `NULL` by default, and be converted to ClickHouse type `Array<Nullable>` eventually.

PR #82, opened 2022-04-14
Support partitioned by Date Type

camper42:
unquoted partition condition may cause error like below when partition column type is Date
`xenon.clickhouse.exception.ClickHouseServerException: [43] DB::Exception: Illegal types of arguments (Date, Int64) of function equals: While processing load_date = ((2022 - 4) - 12)`

### DIFF
--- a/spark-3.2/clickhouse-spark/src/main/scala/xenon/clickhouse/read/InputPartitions.scala
+++ b/spark-3.2/clickhouse-spark/src/main/scala/xenon/clickhouse/read/InputPartitions.scala
@@ -40,6 +40,11 @@ case class ClickHouseInputPartition(
 
   def partFilterExpr: String = partition match {
     case NoPartitionSpec => "1=1"
-    case PartitionSpec(part, _, _) => s"${table.partition_key} = $part"
+    case PartitionSpec(part, _, _) => (part.contains("-"), part.contains("(")) match {
+      // quote when partition by a single Date Type column to avoid illegal types of arguments (Date, Int64)
+      case (true, false) => s"${table.partition_key} = '$part'"
+      // Date type column is quoted if there are multi partition columns
+      case _ => s"${table.partition_key} = $part"
+    }
   }
 }
--- a/spark-3.2/clickhouse-spark-it/src/test/scala/xenon/clickhouse/single/ClickHouseSingleSuite.scala
+++ b/spark-3.2/clickhouse-spark-it/src/test/scala/xenon/clickhouse/single/ClickHouseSingleSuite.scala
@@ -14,6 +14,7 @@
 
 package xenon.clickhouse.single
 
+import java.sql.Date
 import java.sql.Timestamp
 
 import org.apache.spark.sql.types._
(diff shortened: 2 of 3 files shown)


camper42:
installing docker on my machine...

pan3793:
I'm afraid the case is a little bit complex, you may need to test with different types and multi cols or exprs cases.

pan3793:
New unit tests is recommended but not enforced in this time, as long as your changes do not break the existing tests.

camper42:
only checked the error my colleague encountered was fix 
writing unit tests to make sure it's really fixed and hasn't broken anything else🙈

spark-3.2/clickhouse-spark/src/main/scala/xenon/clickhouse/read/InputPartitions.scala
```diff
@@ -40,6 +40,11 @@ case class ClickHouseInputPartition(
 
   def partFilterExpr: String = partition match {
     case NoPartitionSpec => "1=1"
-    case PartitionSpec(part, _, _) => s"${table.partition_key} = $part"
+    case PartitionSpec(part, _, _) => (part.contains("-"), part.contains("(")) match {
+      // quote when partition by a single Date Type column to avoid illegal types of arguments (Date, Int64)
```
pan3793, line 44: It's enough for the `Date` case, and we can imporve it in future by extract a method like `compilePartValue`

pan3793 (approved):

pan3793:
Thanks, merging to master and branch-0.1, will release 0.1.2 soon.

spark-3.2/clickhouse-spark/src/main/scala/xenon/clickhouse/read/InputPartitions.scala
camper42: clickhouse 21.6 above add virtual column `_partition_id`, perhaps more suitable for `InputPartitions` ?

pan3793: We need to know the partition value for pruning

pan3793: After second thought, it should be a good idea, because that partition pruning ocuurs before inferring `InputPartition`

pan3793: Compare to the current implementation, switching to `_partition_id` does not have side-effect except for compatibility, we can do that and introduce a configure in case the user uses v21.3

camper42: `planInputPartitionsById` and default to `true` ?

pan3793: SGTM.

PR #85, opened 2022-04-15
Upgrade ClickHouse 22.3.3.44

pan3793:
Docker image supports ARM64


pan3793:
v22.3.3.44 may has issue on ARM64
https://github.com/ClickHouse/ClickHouse/issues/36286

pan3793:
@martin-g The ClickHouse 22.3.3.44 passed integration test on x86_64, would you please check if the integration test works on ARM64?

martin-g:
The problem seems to be in the ARM64 image itself.
On x86_64 I can do:
1) docker run --rm -it clickhouse/clickhouse-server:22.3.3.44 bash
2) root@5401f1d253f9:/# su-exec test
Usage: su-exec user-spec command [args]

But on aarch64 it fails with segfault:
```
docker run --rm -it clickhouse/clickhouse-server:22.3.3.44 bash
/entrypoint.sh: line 39:    10 Segmentation fault      $gosu test -f "$CLICKHOUSE_CONFIG" -a -r "$CLICKHOUSE_CONFIG"
Configuration file '/etc/clickhouse-server/config.xml' isn't readable by user with id '101'
```

To enter in the docker container I need to override the entrypoint:
```
docker run --rm -it -v --entrypoint bash  clickhouse/clickhouse-server:22.3.3.44
```
and then trying to use `su-exec` segfaults:
```
 su-exec test
Segmentation fault
```

pan3793:
Sounds like another issue than https://github.com/ClickHouse/ClickHouse/issues/36286 of ClickHouse, would you mind report it to the ClickHouse? We can upgrade the version it once the upstream releases the patched version.

martin-g:
@pan3793 https://github.com/ClickHouse/ClickHouse/issues/36293

PR #203, opened 2022-11-09
Spark: Allow setting arbitrary options for ClickHouseClient

pan3793:
https://github.com/housepower/spark-clickhouse-connector/issues/201#issuecomment-1308198643


spark-3.3/clickhouse-spark/src/main/scala/xenon/clickhouse/Constants.scala
```diff
@@ -19,14 +19,15 @@ object Constants {
...
-  final val CATALOG_PROP_GRPC_PORT = "grpc_port"
-  final val CATALOG_PROP_HTTP_PORT = "http_port"
-  final val CATALOG_PROP_PROTOCOL  = "protocol"
-  final val CATALOG_PROP_USER      = "user"
-  final val CATALOG_PROP_PASSWORD  = "password"
-  final val CATALOG_PROP_DATABASE  = "database"
-  final val CATALOG_PROP_TZ        = "timezone" // server(default), client, UTC+3, Asia/Shanghai, etc.
+  final val CATALOG_PROP_HOST          = "host"
+  final val CATALOG_PROP_GRPC_PORT     = "grpc_port"
+  final val CATALOG_PROP_HTTP_PORT     = "http_port"
```
cevenrokt, line 24: would you mind to also standard `tcp_port` since `NodeSpec` expects is specifically?

clickhouse-core/src/main/scala/xenon/clickhouse/client/NodeClient.scala
```diff
@@ -34,14 +34,14 @@ object NodeClient {
...
   private val node: ClickHouseNode = ClickHouseNode.builder()
+    .options(nodeSpec.options)
     .host(nodeSpec.host)
     .port(nodeSpec.protocol, nodeSpec.port)
     .database(nodeSpec.database)
     .credentials(ClickHouseCredentials.fromUserAndPassword(nodeSpec.username, nodeSpec.password))
     .build()
 
   private val client: ClickHouseClient = ClickHouseClient.builder()
-    .option(ClickHouseClientOption.ASYNC, false)
```
cevenrokt, line 44: did you mean to remove it?

spark-3.3/clickhouse-spark/src/main/scala/xenon/clickhouse/Constants.scala
pan3793: The https://github.com/ClickHouse/clickhouse-jdbc does not support TPC protocol yet

clickhouse-core/src/main/scala/xenon/clickhouse/client/NodeClient.scala
pan3793: User can setting it by `spark.sql.catalog.<catalog_name>.option.async`

I'm still thinking how to make the option overwriting logic meaningful, you can compile and test if this approach works first.

cevenrokt:
Would these options accept SSL, or is extra work required to use https when specifying `HTTP` protocol?

cevenrokt (approved):

pan3793:
It should support all keys defined in `com.clickhouse.client.config.ClickHouseClientOption`

PR #211, opened 2023-01-31
Build: Bump clickhouse-java 0.4.5

pan3793:
ClickHouse official Java client 0.4.5 has been released a few weeks ago, which brings breaking changes.

https://github.com/ClickHouse/clickhouse-jdbc/releases/tag/v0.4.5

Notes: 0.4.0 ~ 0.4.4 does not work, because of the following patch which is included in 0.4.5

And there is a known issue on gRPC w/ lz4, disable it then.

> - ClickHouseResponse.getInputStream may return closed input stream
> - https://github.com/ClickHouse/clickhouse-java/pull/1327


pan3793:
Thanks @@zhicwu, the HTTP protocol looks good now, looks there still something wrong on gRPC

PR #218, opened 2023-02-16
Allow dropping partitions with Date type column

tangyl:
This pull request enables the `DROP PARTITION` operation to work with tables that have a column of type Date as partition. With this change, you can use a statement like the following to drop a partition from a ClickHouse table:

```
ALTER TABLE <table> DROP PARTITION (dt='2023-02-15')
```
where dt is a column with type Date. Before this change, attempting to drop a partition with a Date column would result in an error.

I've tested this change by running the `DROP PARTITION` operation on a table with a Date column, and I've verified that it works as expected.

Let me know if you have any questions or concerns about this change.


pan3793:
Would you like to supply a unit test as well?

tangyl:
@pan3793 is unit test mandatory? Seems it would take more efforts.

pan3793:
OK, looks pretty safe, merged.

Thanks for your contribution.

PR #223, opened 2023-02-28
Spark: Support `tcp_port` in catalog option

camper42:


pan3793:
Thanks for fixing this, does the upstream ClickHouse Java Client support TCP protocol now?

camper42:
> clickhouse-cli-client(wrapper of ClickHouse native command-line client) was added in 0.3.2-patch10, clickhouse-tcp-client will be available in 0.5

from 0.5.0, we use 0.3.2-patch11, it is supported

camper42:
> clickhouse-cli-client(wrapper of ClickHouse native command-line client) was added in 0.3.2-patch10, clickhouse-tcp-client will be available in 0.5

if use tcp, we need a clickhouse cli in our environment before clickhouse-java 0.5

camper42:
we have performance issue when use http protocol.

we set batchSize to 1000 to avoid `xenon.clickhouse.exception.CHServerException: [HTTP]user@host:8123}/default [1002] Error writing request body to server`, but it's 2x slower vs 0.4.0 connect+grpc.

we want to give tcp a try

pan3793 (approved):

PR #233, opened 2023-04-27
Core: Deprecate gRPC protocol

pan3793:
Due to poor support from upstream

https://github.com/ClickHouse/clickhouse-java/issues/1300#issuecomment-1483983951

> gRPC is experimental and not suitable for production. It's reserved only for testing purpose and I'll drop it someday.


pan3793:
cc @zhicwu, as a downstream project of [ClickHouse/clickhouse-java](https://github.com/ClickHouse/clickhouse-java), I decide to deprecate the gRPC too.

PR #237, opened 2023-05-09
Spark 3.4: Adapt to SPARK-43390 to allow reserving schema nullability on CTAS/RTAS

pan3793:
Fix https://github.com/housepower/spark-clickhouse-connector/issues/167
Fix https://github.com/housepower/spark-clickhouse-connector/issues/236


pan3793:
@ivanychev @sketchmind The issue about losing schema nullability on CTAS/RTAS was resolved by SPARK-43390 upstream, this change is to adapt it on the connector side, but since it only will be available in Spark 3.5.0, you may need to manually cherry-pick the patch to your internal Spark branch to achieve this capability.

This change won't be verified via GitHub CI, would be great if you can share your testing result.

jonasasx:
The method `def useNullableQuerySchema` has been overridden in the class `com.clickhouse.spark.ClickHouseTable`. However, the override should be implemented in `com.clickhouse.spark.ClickHouseCatalog`. This is because `org.apache.spark.sql.connector.catalog.TableCatalog` includes the following **default** method signature:


```
  default boolean useNullableQuerySchema() {
    return true;
  }
```

PR #253, opened 2023-06-25
Support writing with functions in distribute/partition expressions

Yxang:
## What does this PR do?
This PR introduces support for writing to ClickHouse tables with functions in distribute/partition expressions after Spark 3.4's merge of [SPARK-39607](https://issues.apache.org/jira/browse/SPARK-39607), with the need to implement ClickHouse functions into corresponding Spark ones. 

The implemented functions in this PR include:
- Time-related functions: `toYear`, `toYYYYMM`,  and `toYYYYMMDD` (with only Spark `date` input), and `toHour` (with only Spark `timestamp` or `timestamp_ntz`) 
   - original `Transform` like [`YearsTransform`](https://github.com/apache/spark/blob/v3.4.0/sql/catalyst/src/main/scala/org/apache/spark/sql/connector/expressions/expressions.scala#L242) is not evaluable
- Hash functions: `murmurHash3_64`, `murmurHash3_32`, `murmurHash2_64`, `murmurHash2_32`, and `cityHash64` with maximum of 5 `string` arguments.

Also existing v2 function for `xxHash64` is available.

Corresponding tests are also included.

## Limitations

Main limitation is Spark function cannot fully match Clickhouse function's behavior, including:
- Hash functions support for maximum of 5 arguments, limited by Spark's v2 function does not support variable length of inputs
- Hash functions only support `string` data type, due to the complexity in aligning Spark's and ClickHouse's implementation of hash function and type casting
- When input type mismatches with the required type, error message ```org.apache.spark.sql.AnalysisException: some_function(arg) is not currently supported``` is ambiguous and misleading
- Time functions only support one input type. Ideally UDF should support multiple input types.

PR #268, opened 2023-08-25
Spark 3.4: Enhance FunctionRegistry to support more hash functions

pan3793:
This is part of #253, with some enhancements.

Lead-authored-by: Xinyuan Yang <yangxinyuan100@gmail.com>
Co-authored-by: Cheng Pan <chengpan@apache.org>


spark-3.4/clickhouse-spark/src/main/scala/xenon/clickhouse/func/MultiStringArgsHash.scala
```diff
@@ -0,0 +1,62 @@
...
+        override def canonicalName: String = s"clickhouse.$name"
+        override def resultType: DataType = LongType
+        override def toString: String = name
+        override def produceResult(input: InternalRow): Long = {
+          val inputStrings = new Array[Any](input.numFields)
+          var i = 0
+          do {
+            inputStrings(i) = input.getUTF8String(i).getBytes
+            i += 1
+          } while (i < input.numFields)
```
pan3793, line 54: According to https://github.com/databricks/scala-style-guide#traversal-and-zipwithindex, we should use while for performance-sensitive code.

PR #274, opened 2023-10-20
Spark: Cached ClickHouse table is not used by SQL

zhouyifan279:
Run Spark SQL: 
```
cache table clickhosue.db.tbl;
explain select * from clickhosue.db.tbl;
== Physical Plan ==
*(1) Project [i#776]
+- BatchScan[i#776] class xenon.clickhouse.read.ClickHouseBatchScan RuntimeFilters: []
```
From the Physical Plan, we found the cached table was not used by this query.


clickhouse-core/src/main/scala/xenon/clickhouse/spec/TableSpec.scala
```diff
@@ -27,11 +27,11 @@ case class TableSpec(
   uuid: String,
   engine: String,
   is_temporary: Boolean,
-  data_paths: Array[String],
+  data_paths: List[String],
   metadata_path: String,
   metadata_modification_time: LocalDateTime,
-  dependencies_database: Array[String],
-  dependencies_table: Array[String],
```
pan3793, line 34: the key point is equals? if so, does Seq work?

zhouyifan279: Seq is not Serializable

pan3793 (approved):

PR #303, opened 2024-03-26
Array and Map should not be wrapped in Nullable

heartdance:
ClickHouse not support Array or Map wrapped in Nullable.
ex: Array(Nullable(String)) is supported, but Nullable(Array(String)) is not supported.

### DIFF
--- a/spark-3.5/clickhouse-spark/src/main/scala/org/apache/spark/sql/clickhouse/SchemaUtils.scala
+++ b/spark-3.5/clickhouse-spark/src/main/scala/org/apache/spark/sql/clickhouse/SchemaUtils.scala
@@ -70,25 +70,25 @@ object SchemaUtils {
     (catalystType, chColumn.isNullable)
   }
 
-  def toClickHouseType(catalystType: DataType): String =
+  def toClickHouseType(catalystType: DataType, nullable: Boolean): String =
     catalystType match {
-      case BooleanType => "UInt8"
-      case ByteType => "Int8"
-      case ShortType => "Int16"
-      case IntegerType => "Int32"
-      case LongType => "Int64"
-      case FloatType => "Float32"
-      case DoubleType => "Float64"
-      case StringType => "String"
-      case VarcharType(_) => "String"
-      case CharType(_) => "String" // TODO: maybe FixString?
-      case DateType => "Date"
-      case TimestampType => "DateTime"
-      case DecimalType.Fixed(p, s) => s"Decimal($p, $s)"
-      case ArrayType(elemType, nullable) => s"Array(${maybeNullable(toClickHouseType(elemType), nullable)})"
+      case BooleanType => maybeNullable("UInt8", nullable)
+      case ByteType => maybeNullable("Int8", nullable)
+      case ShortType => maybeNullable("Int16", nullable)
+      case IntegerType => maybeNullable("Int32", nullable)
+      case LongType => maybeNullable("Int64", nullable)
+      case FloatType => maybeNullable("Float32", nullable)
+      case DoubleType => maybeNullable("Float64", nullable)
+      case StringType => maybeNullable("String", nullable)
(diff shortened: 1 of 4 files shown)


spark-3.5/clickhouse-spark/src/main/scala/org/apache/spark/sql/clickhouse/SchemaUtils.scala
```diff
@@ -110,7 +110,7 @@ object SchemaUtils {
       }
 
   private[clickhouse] def maybeNullable(chType: String, nullable: Boolean): String =
-    if (nullable) wrapNullable(chType) else chType
+    if (nullable && !chType.startsWith("Array") && !chType.startsWith("Map")) wrapNullable(chType) else chType
```
pan3793, line 113: If possible, prefer using strong data type rather than string for type matching, for this case, better to handle it on the caller's side

pan3793:
Do you have the opportunity to add an e2e test for your change? And you only need to change the code in spark-3.5, there is a script to backport your change to lower Spark versions' module

spark-3.5/clickhouse-spark/src/main/scala/org/apache/spark/sql/clickhouse/SchemaUtils.scala
heartdance: ok

heartdance:
> Do you have the opportunity to add an e2e test for your change? And you only need to change the code in spark-3.5, there is a script to backport your change to lower Spark versions' module

ok, added

spark-3.5/clickhouse-spark/src/main/scala/org/apache/spark/sql/clickhouse/SchemaUtils.scala
```diff
@@ -70,25 +70,25 @@ object SchemaUtils {
...
+      case CharType(_) => maybeNullable("String", nullable) // TODO: maybe FixString?
+      case DateType => maybeNullable("Date", nullable)
+      case TimestampType => maybeNullable("DateTime", nullable)
+      case DecimalType.Fixed(p, s) => maybeNullable(s"Decimal($p, $s)", nullable)
+      case ArrayType(elemType, containsNull) => s"Array(${toClickHouseType(elemType, containsNull)})"
       // TODO currently only support String as key
-      case MapType(keyType, valueType, nullable) if keyType.isInstanceOf[StringType] =>
-        s"Map(${toClickHouseType(keyType)},${maybeNullable(toClickHouseType(valueType), nullable)})"
+      case MapType(keyType, valueType, valueContainsNull) if keyType.isInstanceOf[StringType] =>
+        s"Map(${toClickHouseType(keyType, nullable = false)},${toClickHouseType(valueType, valueContainsNull)})"
```
pan3793, line 91:
```suggestion
        s"Map(${toClickHouseType(keyType, nullable = false)}, ${toClickHouseType(valueType, valueContainsNull)})"
```

spark-3.5/clickhouse-spark/src/test/scala/org/apache/spark/sql/clickhouse/SchemaUtilsSuite.scala
```diff
@@ -199,7 +201,9 @@ class SchemaUtilsSuite extends AnyFunSuite {
       ("id", "Int32", ""),
       ("food", "String", " COMMENT 'food'"),
       ("price", "Decimal(2, 1)", " COMMENT 'price usd'"),
-      ("remark", "Nullable(String)", "")
+      ("remark", "Nullable(String)", ""),
+      ("ingredient", "Array(Nullable(String))", ""),
+      ("nutrient", "Map(String,Nullable(String))", "")
```
pan3793, line 206:
```suggestion
      ("nutrient", "Map(String, Nullable(String))", "")
```

pan3793 (approved):

PR #318, opened 2024-06-21
replace master with main branch & license fix

mzitnik:


.github/workflows/build-and-test.yml
```diff
@@ -18,10 +18,12 @@ on:
     branches:
       - "master"
       - "branch-*"
+      - "main"
```
pan3793, line 21: why not just replace "master" with "main", instead of appending?

LICENSE
```diff
@@ -1,3 +1,5 @@
+Copyright 2016-2024 ClickHouse, Inc.
```
pan3793, line 1: this is not the right place to add the copyright, this file should be kept as-is. 

usually, copyright info should be added to each file's license header, and the NOTICE file.

LICENSE
```diff
@@ -186,7 +188,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright 2016-2024 ClickHouse, Inc.
```
pan3793, line 191: ditto, should be kept as-is, you can refer to some ASF projects.

README.md
```diff
@@ -1,7 +1,7 @@
 Spark ClickHouse Connector
 ===
 [![Maven Central](https://maven-badges.herokuapp.com/maven-central/com.github.housepower/clickhouse-spark-runtime-3.3_2.12/badge.svg)](https://search.maven.org/search?q=clickhouse-spark-runtime)
-[![License](https://img.shields.io/github/license/housepower/spark-clickhouse-connector)](https://github.com/housepower/spark-clickhouse-connector/blob/master/LICENSE)
+[![License](https://img.shields.io/github/license/housepower/spark-clickhouse-connector)](https://github.com/housepower/spark-clickhouse-connector/blob/main/LICENSE)
```
pan3793, line 4: org name should be replaced

.github/workflows/build-and-test.yml
mzitnik: I was thinking after we merge to master rename the branch to main and also add protection to the brach (so eventually master will be removed)

LICENSE
mzitnik: This our convention in ClickHouse i will check with legal

LICENSE
mzitnik: same here as i have mention on the previous comment

mshustov (approved):

.github/workflows/deploy-website.yml
```diff
@@ -17,6 +17,7 @@ on:
   push:
     branches:
       - "master"
+      - "main"
```
pan3793, line 20: seems the ClickHouse org does not enable the GitHub Pages, you may need to find out another place to host the docs.

mzitnik: We will probably move it to ClickHouse docs 
Just to verify you are referring to the website

pan3793: yes, the website, the previous link is https://housepower.github.io/spark-clickhouse-connector, is not available after the transfer.

PR #343, opened 2024-07-17
Change package from xenon.clickhouse to com.clickhouse

mzitnik:
## Summary
Change package from xenon.clickhouse to com.clickhouse

## Checklist
Delete items not relevant to your PR:
- [ ] Unit and integration tests covering the common scenarios were added
- [ ] A human-readable description of the changes was provided to include in CHANGELOG


clickhouse-core/src/main/scala/com/clickhouse/io/ObservableOutputStream.scala
```diff
@@ -12,10 +12,9 @@
  * limitations under the License.
  */
 
-package xenon.clickhouse.io
-
-import xenon.clickhouse.Utils
+package com.clickhouse.io
 
+import com.clickhouse.Utils
```
pan3793, line 17: will this conflict with ClickHouse Java client? Do we need a sub-package?

pan3793:
> ... to `com.clickhouse`

may be to `com.clickhouse.spark`, to avoid conflicts with the Java client

clickhouse-core/src/main/scala/com/clickhouse/io/ObservableOutputStream.scala
mzitnik: Good point 
Thanks

mshustov (approved):

PR #377, opened 2024-12-21
Support ClickHouse insert settings for table writes

### LINKED ISSUE #369: Support to set ClickHouse insert settings when writing to a table
We can add custom settings to ClickHouse SQL when running an insert. 

Example:
`INSERT INTO [table] SETTINGS [...]`

Currently, there is no way to achieve this with the spark connector.

Maybe spark-clickhouse-connector can read settings starting with `spark.clickhouse.write.settings`. and apply them when write into clickhouse tables.

mahdimalverdi:
## Summary
allow write with settings.

close #369


mahdimalverdi:
Some problems

PR #427, opened 2025-10-13
Detect runtime env

### LINKED ISSUE #397: Add runtime environment info (e.g. Glue, Databricks) to User-Agent string
### Description

For debugging and monitoring purposes, it would be helpful to include details about the runtime environment in the User-Agent string sent by the Spark ClickHouse connector when communicating with ClickHouse servers.

For example, if the job is running on:
* AWS Glue
* Databricks
* EMR - (future scope)
* Kubernetes - (future scope)
* Local Spark - (future scope)

### Implementation ideas

* Try to detect the environment from common Spark configs (spark.databricks.clusterUsageTags.clusterOwner, spark.glue.source, etc.).
* Try to infer the runtime environment by iterating through the stack classes.

BentsiLeviav:
Only Glue (AWS) and Databricks environments were tested and verified; the rest of the managed services were added based on online research and vendor documentation. 

close #397

### DIFF
--- a/clickhouse-core/src/main/scala/com/clickhouse/spark/client/NodeClient.scala
+++ b/clickhouse-core/src/main/scala/com/clickhouse/spark/client/NodeClient.scala
@@ -25,10 +25,10 @@ import com.clickhouse.spark.format.{
   NamesAndTypes,
   SimpleOutput
 }
+import com.clickhouse.spark.Utils.RuntimeDetector.detectRuntime
 import com.clickhouse.spark.spec.NodeSpec
 import com.fasterxml.jackson.databind.JsonNode
 import com.fasterxml.jackson.databind.node.ObjectNode
-import com.clickhouse.spark.format._
 
 import java.io.InputStream
 import java.util.UUID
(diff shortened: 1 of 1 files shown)


clickhouse-core/src/main/scala/com/clickhouse/spark/Utils.scala
```diff
@@ -186,4 +187,153 @@ object Utils extends Logging {
...
+      RuntimeDetector.detectViaStackTrace()
+        .orElse(RuntimeDetector.detectViaClassLoader())
+        .orElse(RuntimeDetector.detectViaThreadNames())
+
+    /**
+     * Examines the current stack trace and loaded classes for platform-specific signatures
+     */
+    def detectViaStackTrace(): Option[String] = {
+      val stackTrace = Thread.currentThread().getStackTrace
+      val stackClasses = stackTrace.map(_.getClassName.toLowerCase)
```
mzitnik, line 203: I will convert the map to a set and just make exists and looks like it can applied to other detections

clickhouse-core/src/main/scala/com/clickhouse/spark/client/NodeClient.scala
```diff
@@ -42,23 +42,42 @@ class NodeClient(val nodeSpec: NodeSpec) extends AutoCloseable with Logging {
...
-      val versions = version.split("_")
-      if (versions.length < 3) {
-        "Spark-ClickHouse-Connector"
-      } else {
-        val sparkVersion = versions(0)
-        val scalaVersion = versions(1)
-        val connectorVersion = versions(2)
-        s"${title}/${connectorVersion} (fv:spark/${sparkVersion}, lv:scala/${scalaVersion})"
-      }
+    buildUserAgent(title, version)
```
mzitnik, line 48: Did we validate that this once, only once, on every node? Since it can be a heavy operation

clickhouse-core/src/main/scala/com/clickhouse/spark/Utils.scala
BentsiLeviav: As we discussed off-line, we do pattern matching and not search for exact term, this is why the proposed change is not relevant.

clickhouse-core/src/main/scala/com/clickhouse/spark/client/NodeClient.scala
BentsiLeviav: Yes, it is being triggered once for any node client initiation. It is important to mention that multiple node clients are being created for a single application running a simple query, but we made sure it happens once for any initiation.

mzitnik: Ok, I think we are okay. I would rather have it initialized once per JVM. (but we can handle it later)

PR #452, opened 2025-11-16
Spark 4 support

ShimonSte:
# Add Spark 4.0 Support

## Overview
This PR adds support for Apache Spark 4.0 with Scala 2.13 to the ClickHouse Spark Connector.

### Core Implementation
- **Spark 4.0 Module**: Added `spark-4.0/` directory with connector implementation and integration tests
- **Scala 2.13**: Full compatibility with Scala 2.13 (Spark 4.0 requirement)
- **Java 17**: Updated build configuration to use Java 17 for Spark 4.0

## Compatibility
- **Spark 4.0.x** with Scala 2.13
- **Java 17** (minimum requirement for Spark 4.0)
- Maintains backward compatibility with Spark 3.3, 3.4, and 3.5

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -49,8 +49,8 @@ project.ext {
     spark_prefix = "spark_${spark_binary_version.replace('.', '')}"
     scala_prefix = "scala_${scala_binary_version.replace('.', '')}"
 
-    scala_212_version = project.getProperty("${spark_prefix}_scala_212_version")
-    scala_213_version = project.getProperty("${spark_prefix}_scala_213_version")
+    scala_212_version = project.findProperty("${spark_prefix}_scala_212_version") ?: "2.12.18"
+    scala_213_version = project.findProperty("${spark_prefix}_scala_213_version") ?: "2.13.8"
     scala_version = project.getProperty("${scala_prefix}_version")
 
     antlr_version = project.getProperty("${spark_prefix}_antlr_version")


.github/workflows/build-and-test.yml
```diff
@@ -58,7 +65,7 @@ jobs:
       - uses: actions/setup-java@v4
         with:
           distribution: zulu
-          java-version: 8
+          java-version: 17
```
mzitnik: I think we can merge `run-tests` and `run-tests-with-specific-clickhouse` into one test. Do you know what version we are running on `run-tests`

.github/workflows/cloud.yml
```diff
@@ -34,8 +34,21 @@ jobs:
       max-parallel: 1
       fail-fast: false
       matrix:
-        spark: [ 3.3, 3.4, 3.5 ]
-        scala: [ 2.12, 2.13 ]
+        spark: [ '3.3', '3.4', '3.5', '4.0' ]
+        scala: [ '2.12', '2.13' ]
+        java: [ 8 ]
```
mzitnik: why not `java: [ 8, 17 ]`

.github/workflows/tpcds.yml
```diff
@@ -30,14 +30,27 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        spark: [ 3.3, 3.4, 3.5 ]
-        scala: [ 2.12, 2.13 ]
+        spark: [ '3.3', '3.4', '3.5', '4.0' ]
+        scala: [ '2.12', '2.13' ]
+        java: [ 8 ]
```
mzitnik: same here `java: [ 8, 17 ]`

build.gradle
```diff
@@ -106,7 +106,11 @@ allprojects {
 subprojects {
     apply plugin: "scala"
     apply plugin: "java-library"
-    apply plugin: "org.scoverage"
+    // Disable scoverage when running Metals' bloopInstall to avoid plugin resolution issues
+    def isBloopInstall = gradle.startParameter.taskNames.any { it.contains('bloopInstall') }
```
mzitnik, line 110: Did you check if it does not affect IntelliJ?

clickhouse-core/src/testFixtures/scala/com/clickhouse/spark/base/ClickHouseSingleMixIn.scala
```diff
@@ -80,4 +84,20 @@ trait ClickHouseSingleMixIn extends AnyFunSuite with ForAllTestContainer with Cl
   override def clickhousePassword: String = CLICKHOUSE_PASSWORD
   override def clickhouseDatabase: String = CLICKHOUSE_DB
   override def isSslEnabled: Boolean = false
+
+  override def beforeAll(): Unit = {
+    val startTime = System.currentTimeMillis()
+    println(s"[ClickHouseSingleMixIn] Starting ClickHouse container: $CLICKHOUSE_IMAGE")
+    super.beforeAll() // This starts the container and makes mappedPort available
+    val duration = System.currentTimeMillis() - startTime
+    println(
```
mzitnik: maybe we should use log instead of println

.github/workflows/build-and-test.yml
ShimonSte: The `run-tests` run on 23.8 (the default version in the `ClickHouseSingleMixIn` file).
We could merge them into one test but then we'll need to update the matrix that would run all tests on all CH versions (might actually be a good idea but it'll create a lot of new tests. maybe we should change this only when we fix the tests so that they will be able to run concurrently)

.github/workflows/cloud.yml
ShimonSte: Again this would cause many more versions of the tests to run (today it is only running 8).

.github/workflows/tpcds.yml
ShimonSte: Same as previous comment

build.gradle
ShimonSte: It shouldn't affect as IntelliJ uses its own build server and it doesn't use Bloop at all

.github/workflows/build-and-test.yml
mzitnik: In general, this test is against local ClickHouse, not the cloud, so it does not have the concurrency issue

.github/workflows/cloud.yml
mzitnik: I'am ok with that since we might have a so feature that is not supported in 8 that can break so it better to test

spark-4.0/clickhouse-spark/src/main/scala/org/apache/spark/sql/clickhouse/SparkUtils.scala
```diff
@@ -25,7 +25,8 @@ object SparkUtils {
 
   lazy val MAJOR_MINOR_VERSION: (Int, Int) = VersionUtils.majorMinorVersion(SPARK_VERSION)
 
-  def toArrowSchema(schema: StructType, timeZoneId: String): Schema = ArrowUtils.toArrowSchema(schema, timeZoneId, true)
+  def toArrowSchema(schema: StructType, timeZoneId: String): Schema =
+    ArrowUtils.toArrowSchema(schema, timeZoneId, true, false)
```
mzitnik, line 29: looking at the API https://spark.apache.org/docs/latest/api/java/org/apache/spark/sql/util/ArrowUtils.html#toArrowSchema(org.apache.spark.sql.types.StructType,java.lang.String,boolean,boolean)

Do we know how this `boolean largeVarTypes = false` will affect us?

ShimonSte: largeVarTypes = false uses 32-bit offsets (max 2GB per string/binary value) which is the Spark 3.x behavior, while true would use 64-bit offsets for massive values but likely break compatibility.

gradle.properties
```diff
@@ -29,26 +29,32 @@ clickhouse_client_v2_version=0.9.4
 spark_33_version=3.3.4
 spark_34_version=3.4.2
 spark_35_version=3.5.1
+spark_40_version=4.0.1
 
 spark_33_scala_212_version=2.12.15
 spark_34_scala_212_version=2.12.17
 spark_35_scala_212_version=2.12.18
+spark_40_scala_212_version=2.12.18
```
pan3793: spark 4.0 does not support scala 2.12

gradle.properties
pan3793: the apache commons libs have good backward compatibility, let's upgrade them to align with latest stable Spark version.

pan3793, line 70: also suggest upgrading scalatest and flexmark to align with spark 4.0.1

.github/workflows/tpcds.yml
```diff
@@ -30,14 +30,22 @@ jobs:
...
-        spark: [ 3.3, 3.4, 3.5 ]
-        scala: [ 2.12, 2.13 ]
+        spark: [ '3.3', '3.4', '3.5', '4.0' ]
+        scala: [ '2.12', '2.13' ]
+        java: [ 8, 17 ]
+        exclude:
+          # Spark 4.0 only supports Scala 2.13
+          - spark: '4.0'
+            scala: '2.12'
+          # Spark 4.0 requires Java 11+
```
pan3793:
```suggestion
          # Spark 4.0 requires Java 17+
```

gradle.properties
ShimonSte: Good catch

gradle.properties
ShimonSte: We should keep flexmark at the current version as newer versions do not support java 8.
I think it is time we start talking about deprecating java 8, I'll raise it up as an issue.

pan3793: Unfortunately, in the big data world, especially the Hadoop ecosystem still heavily stuck with Java 8, I would suggest deferring such deprecation until we hit significant issues.

Additional notes: Spark is not a service, but more like a library or programming framework that mixes user code at runtime, so it's impossible to make aggressive upgrades and ask all users to tune all their jobs. In practice, some Spark jobs may be unable to upgrade to a new Spark version in their lifecycle.

ShimonSte: Great point. Makes sense to keep java as-is for now. Thanks for the context!

pan3793 (approved):
overall LGTM, (I didn't go through all files, and I suppose most scala files are just copy-pasted, maybe with a little tuning to pass compile.)

mzitnik (approved):

PR #453, opened 2025-11-20
Feature/struct type support

ShimonSte:
## Add StructType Support for ClickHouse Tuple Types

This PR adds full support for mapping ClickHouse `Tuple` types to Spark `StructType`, enabling read and write operations for structured data.

### Features

- **Type Mapping**: Bidirectional conversion between ClickHouse Tuple and Spark StructType
  - Named tuples: `Tuple(name String, age Int32)` ↔ `StructType(StructField("name", StringType), ...)`
  - Unnamed tuples: `Tuple(String, Int32)` ↔ `StructType(StructField("_1", StringType), ...)`
- **Boolean Support**: Correctly map ClickHouse `Bool` to Spark `BooleanType` (previously mapped to `ShortType`)
- **Format Support**: Works with all read/write formats (JSON, Binary, Arrow)
- **Nested Structures**: Full support for nested tuples and complex types

### DIFF
--- a/spark-3.3/clickhouse-spark-it/src/test/scala/org/apache/spark/sql/clickhouse/single/ClickHouseWriterTestBase.scala
+++ b/spark-3.3/clickhouse-spark-it/src/test/scala/org/apache/spark/sql/clickhouse/single/ClickHouseWriterTestBase.scala
@@ -143,14 +143,14 @@ trait ClickHouseWriterTestBase extends SparkClickHouseSingleTest {
 
       val result = spark.table("test_db.test_write_bool_null").orderBy("id").collect()
       assert(result.length == 3)
-      // Boolean is stored as UInt8 in ClickHouse, reads back as Short
-      assert(result(0).getShort(1) == 1)
+      // Boolean is stored as Bool in ClickHouse, reads back as Boolean
+      assert(result(0).getBoolean(1) == true)
       assert(result(1).isNullAt(1))
-      assert(result(2).getShort(1) == 0)
+      assert(result(2).getBoolean(1) == false)
     }
   }
 
-  // NOTE: ClickHouse stores Boolean as UInt8, so it reads back as Short (0 or 1)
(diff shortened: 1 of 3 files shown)


spark-4.0/clickhouse-spark/src/main/scala/com/clickhouse/spark/read/format/ClickHouseBinaryReader.scala
```diff
@@ -147,6 +147,35 @@ class ClickHouseBinaryReader(
...
           }
         ArrayBasedMapData(convertedMap)
+      case struct: StructType =>
+        // ClickHouse Java client can return tuples as either:
+        // - Array[Object] for unnamed tuples
+        // - util.List[Object] for named tuples
+        val fieldValues = value match {
+          case arr: Array[Object] =>
+            if (arr.length != struct.fields.length) {
+              throw CHClientException(
```
mzitnik, line 157: Did we see this case happening?

mzitnik (commented):
We are testing heavily against the case that we first created the schema using Spark Write, and then read should we also test if some one created the schema before and we only read using Apache Spark

spark-4.0/clickhouse-spark/src/main/scala/com/clickhouse/spark/read/format/ClickHouseBinaryReader.scala
ShimonSte: Regarding the tests, they exist (look [here](https://github.com/ClickHouse/spark-clickhouse-connector/pull/453/files#diff-c6e3bc41411111f1ca501752058c6460bdf6f85958298f330bb46c4cd466d473R1334))
Regarding the exception, i did not see it in my tests. its a safe guard.

mzitnik: i was talking mainly on the exception

Issue #404, opened 2025-08-05
Spark 4.0 - Add VARIANT data type support


BentsiLeviav:
[[SPARK-45827] Add VARIANT data type](https://issues.apache.org/jira/browse/SPARK-45827) introduces a new semi-structured data type in Spark 4.0 that can store JSON-like data with dynamic schemas, similar to ClickHouse's JSON columns or Variant types.

PR #456, opened 2025-11-25
Feature/variant type support

ShimonSte:
### Summary
Implements full support for Spark 4.0's `VariantType` mapped to ClickHouse's `JSON` type, enabling efficient storage and querying of JSON data. Addresses #404 

### Changes
- **Schema Mapping**: Map ClickHouse `JSON`  <-> Spark `VariantType`
- **Binary Reader**: Handle Java collections from ClickHouse binary protocol
- **JSON Reader**: Parse JSON data into Spark's internal `VariantVal` representation
- **Comprehensive Tests**: 14 test cases covering read/write operations, nested structures, arrays, nulls, and edge cases
- **ClickHouse 25.3**: Upgrade to support native JSON type

### Compatibility
- **Spark Version**: 4.0+
- **ClickHouse Version**: 25.3+
- **Formats**: 
   read -> JSON | Binary
   write -> only JSON (Arrow not supported yet)

### DIFF
--- a/spark-4.0/clickhouse-spark-it/src/test/scala/org/apache/spark/sql/clickhouse/single/ClickHouseWriterTestBase.scala
+++ b/spark-4.0/clickhouse-spark-it/src/test/scala/org/apache/spark/sql/clickhouse/single/ClickHouseWriterTestBase.scala
@@ -1527,12 +1527,9 @@ trait ClickHouseWriterTestBase extends SparkClickHouseSingleTest {
     }
   }
 
-  // TODO: Re-enable this test once Spark fixes NULL VariantType serialization with variant_types
-  // Issue: Spark's codegen calls getValue() on null VariantVal when serializing NULL VariantType
-  // values with variant_types option, causing NullPointerException
-  ignore("write VariantType with variant_types - NULL values with mixed types") {
+  test("write VariantType with variant_types - comprehensive mixed types") {
     val db = "test_db"
-    val tbl = "test_variant_write_mixed_nulls"
+    val tbl = "test_variant_write_comprehensive_mixed"
 
     try {
       spark.sql(s"CREATE DATABASE IF NOT EXISTS $db")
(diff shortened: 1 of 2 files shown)


mzitnik:
@ShimonSte, it's not clear 
Formats: JSON | Binary
read -> json/binary 
write -> only JSON

spark-4.0/clickhouse-spark/src/main/scala/org/apache/spark/sql/clickhouse/SchemaUtils.scala
```diff
@@ -29,7 +29,8 @@ object SchemaUtils extends SQLConfHelper {
     val catalystType = chColumn.getDataType match {
       case Nothing => NullType
       case Bool => BooleanType
-      case String | JSON | UUID | Enum8 | Enum16 | IPv4 | IPv6 => StringType
+      case JSON => VariantType
```
pan3793: maybe we should have a config to allow users to switch back to the old behavior

```
spark.clickhouse.read.jsonAs=[string|variant]
```

spark-4.0/clickhouse-spark/src/main/scala/com/clickhouse/spark/read/format/ClickHouseBinaryReader.scala
```diff
@@ -32,15 +32,31 @@ import com.clickhouse.data.value.{
...
 import java.io.InputStream
 import java.time.{LocalDate, ZoneOffset, ZonedDateTime}
 import java.util
+
+object ClickHouseBinaryReader {
+  private val jsonMapper = new ObjectMapper()
+  private val jsonFactory = new com.fasterxml.jackson.core.JsonFactory()
+
+  private def buildVariantFromValue(value: Any): org.apache.spark.unsafe.types.VariantVal = {
+    val jsonString = jsonMapper.writeValueAsString(value)
```
mzitnik, line 57: should we use try/catch in case we can not parse it

spark-4.0/clickhouse-spark-it/src/test/scala/org/apache/spark/sql/clickhouse/single/ClickHouseReaderTestBase.scala
```diff
@@ -1585,4 +1585,31 @@ trait ClickHouseReaderTestBase extends SparkClickHouseSingleTest {
     }
   }
 
+  test("decode VariantType - read JSON as string with config") {
```
mzitnik, line 1618: Maybe we should also have negative test

spark-4.0/clickhouse-spark/src/main/scala/com/clickhouse/spark/read/format/ClickHouseBinaryReader.scala
ShimonSte: We are writing the Binary Variant value from CH as a string and then loading it as a spark Variant.
If it fails here it should through an error (In our case it would fail the task running not the whole spark application)

mzitnik (approved):

Issue #470, opened 2025-12-04
Implement the TableProvider API


BentsiLeviav:
implementing the TableProvider interface as defined in Spark’s DataSourceV2.

### Focus on:

* DefaultSource entry point

* Integration with existing `Table`, `SupportsWrite`, and `WriteBuilder` logic

* Compatibility with batch write operations

### Resources
* [Java docs](https://spark.apache.org/docs/3.5.4/api/java/org/apache/spark/sql/connector/catalog/TableProvider.html)
* [Interface coverage](https://jaceklaskowski.gitbooks.io/mastering-spark-sql/content/spark-sql-data-source-api-v2.html)

PR #471, opened 2025-12-08
Add TableProvider implementation with predicate pushdown and column p…

ShimonSte:
## Add TableProvider for Format-Based API Access #470 

Implements `ClickHouseTableProvider` to enable `.format("clickhouse")` syntax, essential for **Databricks Unity Catalog** environments where custom catalogs cannot be registered.

#### Core Functionality
- **Format name**: `"clickhouse"` - standard Spark DataSource V2 format registration
- **Predicate pushdown** & **column pruning** - verified via `system.query_log` inspection
- **Auto schema inference** - reads table schema from ClickHouse with ORDER BY generation
- **Nullable key support** - automatic `allow_nullable_key` setting for partition/sorting keys

#### Write Modes
- `SaveMode.Append` - append data to existing table
- `SaveMode.Overwrite` - truncate table before writing (full table overwrite) 

#### Overwrite Implementation
- Implements `SupportsOverwrite` trait in `ClickHouseWriteBuilder`  #33 
- Executes `TRUNCATE TABLE` before writing in overwrite mode
- Optimized distributed table handling with single `ON CLUSTER` command
- **Note**: Dynamic partition overwrite not yet implemented - currently only full table truncate

#### ClickHouse Cloud Compatibility
- **SharedMergeTree support** - added `SharedMergeTree` engine parsing in `AstVisitor`
- **Partition expression filtering** - filters unsupported Spark transformations (e.g., `xxHash64` in Spark 3.3) for SharedMergeTree tables
- **Graceful degradation** - keeps supported partition expressions while filtering incompatible ones
- Maintains Spark-side partitioning benefits while avoiding `AnalysisException` errors
- Applied across all Spark versions (3.3, 3.4, 3.5, 4.0)

### Usage

```scala
// Read with format API
val df = spark.read.format("clickhouse")
  .option("host", "localhost")
  .option("database", "default")
  .option("table", "my_table")
  .load()

// Write with overwrite mode
df.write.format("clickhouse")
  .mode(SaveMode.Overwrite)
  .option("host", "localhost")
  .option("database", "default")
  .option("table", "my_table")
  .save()

// Catalog API still available
spark.table("clickhouse.default.my_table")

<!-- CURSOR_SUMMARY -->
---

> [!NOTE]
> Enables format-based access via DataSource V2 for ClickHouse with improved write semantics and engine compatibility.
> 
> - Adds `ClickHouseTableProvider` (format name: `"clickhouse"`) with schema inference, predicate pushdown, and column pruning; registers via `META-INF/services/...DataSourceRegister`
> - Implements overwrite by adding `SupportsOverwrite` to `ClickHouseWrite` and truncating target tables (incl. `ON CLUSTER` for distributed) before write
> - Extends table capabilities with `OVERWRITE_BY_FILTER` and adds SharedMergeTree parsing in `AstVisitor`
> - Filters unsupported partition expressions for `SharedMergeTree` in `WriteJobDescription` to avoid analysis errors
> - Updates/extends IT tests: new `ClickHouseTableProviderSuite`; reader tests compare Catalog vs TableProvider paths; helper API updated to pass `(db, tbl)`
> - Build/doc updates: new `SPARK_VERSION_DIFFERENCES.md`; service files added; Gradle tweaks (mirror repo, ASM 9.6)
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8b162bf21e1f79ea35d3da54ae165e1f705f328f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>
<!-- /CURSOR_SUMMARY -->

### DIFF
--- a/spark-3.3/clickhouse-spark/src/main/scala/com/clickhouse/spark/ClickHouseTableProvider.scala
+++ b/spark-3.3/clickhouse-spark/src/main/scala/com/clickhouse/spark/ClickHouseTableProvider.scala
@@ -205,10 +205,14 @@ class ClickHouseTableProvider extends TableProvider
       Constants.CATALOG_PROP_USER -> optionsMap.getOrElse("user", "default"),
       Constants.CATALOG_PROP_PASSWORD -> optionsMap.getOrElse("password", "")
     ) ++ (
-      if (optionsMap.contains("ssl")) {
-        Map(Constants.CATALOG_PROP_OPTION_PREFIX + "ssl" -> optionsMap("ssl"))
+      if (optionsMap.contains(Constants.CATALOG_PROP_TZ)) {
+        Map(Constants.CATALOG_PROP_TZ -> optionsMap(Constants.CATALOG_PROP_TZ))
       } else {
         Map.empty[String, String]
+      }
+    ) ++ (
+      optionsMap.view.filterKeys(key => key == "ssl" || key == "ssl_mode").toMap.map {
+        case (key, value) => Constants.CATALOG_PROP_OPTION_PREFIX + key -> value
       }
     )
 
--- a/spark-3.3/clickhouse-spark/src/main/scala/com/clickhouse/spark/ClickHouseTableProvider.scala
+++ b/spark-3.3/clickhouse-spark/src/main/scala/com/clickhouse/spark/ClickHouseTableProvider.scala
@@ -128,29 +128,33 @@ class ClickHouseTableProvider extends TableProvider
     val hasOrderBy = props.contains("order_by") || props.contains("local.order_by")
 
     if (!hasOrderBy) {
-      addDefaultOrderBy(schema, props)
-    } else {
-      ensureNullableKeySupport(schema, props)
+      throw new IllegalArgumentException(
+        "Required option 'order_by' is missing when creating a new table. " +
+          "Please provide it via .option('order_by', 'column_name') or .option('order_by', 'col1, col2'). " +
+          "The ORDER BY clause specifies the sorting key for the ClickHouse table."
+      )
     }
-  }
 
-  private def addDefaultOrderBy(schema: StructType, props: Map[String, String]): Map[String, String] = {
-    log.info(s"Schema fields: ${schema.fields.map(f => s"${f.name}(nullable=${f.nullable})").mkString(", ")}")
+    // Validate that order_by columns exist in the schema
+    val orderByKey = props.getOrElse("order_by", props.getOrElse("local.order_by", ""))
+    val orderByColumns = orderByKey
+      .split(",")
+      .map(_.trim)
+      .filter(_.nonEmpty)
 
-    val orderByColumn = schema.fields
-      .find(!_.nullable)
-      .map(_.name)
-      .getOrElse {
-        throw new IllegalArgumentException(
-          "Cannot auto-generate ORDER BY clause: all columns are nullable. " +
-            "ClickHouse Cloud requires non-nullable columns in ORDER BY. " +
-            "Please either: (1) make at least one column non-nullable, or " +
-            "(2) explicitly set 'order_by' option to specify which column(s) to use."
-        )
-      }
+    val schemaFieldNames = schema.fields.map(_.name).toSet
+    val missingColumns = orderByColumns.filterNot(colName =>
+      schemaFieldNames.exists(_.equalsIgnoreCase(colName))
+    )
 
-    log.info(s"Auto-generated ORDER BY clause: $orderByColumn")
-    props + ("order_by" -> orderByColumn)
+    if (missingColumns.nonEmpty) {
+      throw new IllegalArgumentException(
+        s"ORDER BY column(s) not found in schema: ${missingColumns.mkString(", ")}. " +
+          s"Available columns: ${schema.fields.map(_.name).mkString(", ")}"
+      )
+    }
+
+    ensureNullableKeySupport(schema, props)
   }
 
   private def ensureNullableKeySupport(schema: StructType, props: Map[String, String]): Map[String, String] = {
--- a/spark-3.3/clickhouse-spark/src/main/scala/com/clickhouse/spark/ClickHouseTableProvider.scala
+++ b/spark-3.3/clickhouse-spark/src/main/scala/com/clickhouse/spark/ClickHouseTableProvider.scala
@@ -1,0 +1,215 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.clickhouse.spark
+
+import com.clickhouse.spark.client.NodeClient
+import org.apache.spark.internal.Logging
+import org.apache.spark.sql.catalyst.analysis.NoSuchTableException
+import org.apache.spark.sql.connector.catalog.{Identifier, Table, TableProvider}
+import org.apache.spark.sql.connector.expressions.Transform
+import org.apache.spark.sql.sources.DataSourceRegister
+import org.apache.spark.sql.types.StructType
+import org.apache.spark.sql.util.CaseInsensitiveStringMap
+
+import java.util
+
+/**
+ * TableProvider implementation for ClickHouse, enabling format-based access pattern:
+ * {{{
+ *   spark.read
+ *     .format("clickhouse")
+ *     .option("host", "localhost")
+ *     .option("database", "default")
+ *     .option("table", "my_table")
+ *     .load()
+ * }}}
+ *
+ * This implementation delegates to ClickHouseCatalog for all operations,
+ * ensuring consistent behavior between catalog and format APIs.
+ *
+ * Particularly useful for:
+ * - Ad-hoc queries without catalog configuration
+ * - Databricks Unity Catalog environments
+ * - Simple ETL pipelines
+ * - Multi-cluster access patterns
+ */
+class ClickHouseTableProvider extends TableProvider
+    with DataSourceRegister
+    with ClickHouseHelper {
+
+  override def shortName(): String = "clickhouse"
+
+  /**
+   * Infer schema from ClickHouse table.
+   *
+   * Required options:
+   * - host: ClickHouse host
+   * - database: Database name
+   * - table: Table name
+   *
+   * Optional options:
+   * - protocol: http (default), https, grpc
+   * - http_port: HTTP port (default: 8123)
+   * - user: Username (default: "default")
+   * - password: Password (default: empty)
+   * - ssl: Enable SSL (default: false)
+   * - timezone: Timezone (default: "server")
+   */
+  override def inferSchema(options: CaseInsensitiveStringMap): StructType = {
+    log.info(s"Inferring schema for ClickHouse with options: ${options.asCaseSensitiveMap()}")
+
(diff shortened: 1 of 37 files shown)

### CODE REVIEW

File: spark-4.0/clickhouse-spark-it/src/test/scala/org/apache/spark/sql/clickhouse/single/ClickHouseTableProviderSuite.scala [hunk 1]
```diff
@@ -0,0 +1,514 @@
...
+        finalResult.orderBy("id"),
+        Seq(
+          Row(4, "David", 88.7, timestamp("2024-04-05T11:20:00Z")),
+          Row(5, "Eve", 91.2, timestamp("2024-05-12T16:45:00Z"))
+        )
+      )
+    }
+  }
+
+  test("schema inference works correctly") {
```
mzitnik, line 163: For schema inference, we can add more types

mzitnik (commented):
Are we also testing TableProvider in cluster mode (ClickHouse Cluster)?

spark-4.0/clickhouse-spark/src/main/scala/com/clickhouse/spark/write/ClickHouseWrite.scala
```diff
@@ -17,24 +17,58 @@ package com.clickhouse.spark.write
...
+      case _ => false
+    }
+
+    if (hasPartitionFilters) {
+      log.warn(
+        s"Conditional overwrite with partition filters is not fully supported, filters: ${filters.mkString(", ")}"
+      )
+    } else {
+      log.info("Full table overwrite (no partition filters)")
+    }
```
mzitnik: Looks like both messages should be at the same level, why one is `warn` and the other is `info`

spark-4.0/clickhouse-spark/src/main/scala/com/clickhouse/spark/write/ClickHouseWrite.scala
```diff
@@ -17,24 +17,58 @@ package com.clickhouse.spark.write
...
+    with Logging {
 
-  override def build(): Write = new ClickHouseWrite(writeJob)
+  private var isOverwrite: Boolean = false
+  private var overwriteFilters: Array[Filter] = Array.empty
+
+  override def build(): Write = new ClickHouseWrite(writeJob, isOverwrite, overwriteFilters)
+
+  override def overwrite(filters: Array[Filter]): WriteBuilder = {
+    log.info(s"Overwrite mode for table ${writeJob.targetDatabase(false)}.${writeJob.targetTable(false)}")
```
mzitnik, line 43: Should we also add `filters` to logging?

spark-4.0/clickhouse-spark/src/main/scala/com/clickhouse/spark/write/ClickHouseWrite.scala
```diff
@@ -58,10 +92,37 @@ class ClickHouseWrite(
...
+    import com.clickhouse.spark.Utils
+
+    log.info(s"Truncating table ${writeJob.targetDatabase(false)}.${writeJob.targetTable(false)} for overwrite mode")
+
+    Utils.tryWithResource(NodeClient(writeJob.node)) { implicit nodeClient =>
+      writeJob.tableEngineSpec match {
+        case DistributedEngineSpec(_, cluster, local_db, local_table, _, _) =>
+          val sql = s"TRUNCATE TABLE IF EXISTS `$local_db`.`$local_table` ON CLUSTER `$cluster`"
+          log.info(s"Executing: $sql")
+          nodeClient.syncQueryAndCheckOutputJSONEachRow(sql)
```
mzitnik, line 109: i think we have logging in `syncQueryAndCheckOutputJSONEachRow` (We should log there)

spark-4.0/clickhouse-spark/src/main/scala/org/apache/spark/sql/clickhouse/ExprUtils.scala
```diff
@@ -173,9 +173,9 @@ object ExprUtils extends SQLConfHelper with Serializable {
       case Success(t: Transform) => Some(t)
       case Success(_) => None
       case Failure(_) if conf.getConf(IGNORE_UNSUPPORTED_TRANSFORM) => None
-      case Failure(rethrow) => throw new AnalysisException(
-          errorClass = "UNSUPPORTED_FEATURE.TRANSFORM_EXPRESSION",
-          messageParameters = Map("transform" -> rethrow.getMessage),
+      case Failure(rethrow) =>
+        throw CHClientException(
```
mzitnik, line 177: Can you explain why we changed the exception from `AnalysisException` to `CHClientException`

mzitnik (changes requested):
General 
- Let's go over all logging levels and decide the correct level and if we need to log 
- the first comment on extending the coverage 
- Did we solve the conflict with Databricks on Spark 4.0

spark-4.0/clickhouse-spark/src/main/scala/org/apache/spark/sql/clickhouse/ExprUtils.scala
ShimonSte: Spark 4 doesn't have that errorClass (i got an error from the errorClass...)

deploy.gradle
```diff
@@ -81,11 +81,22 @@ subprojects {
                         project.shadow.component(publication)
                         artifact sourceJar
                         artifact javadocJar
+                        
+                        // For databricks variant, set the artifact ID to match the normal runtime (without -databricks)
+                        if (project.name.contains("-databricks")) {
```
mzitnik: We agree eventualy not to publish it maven?

pan3793: Can you guys explain why we need to build a variant for Databricks Spark?

pan3793: OK, I guess Databricks Spark may use a different version of Jackson. In this case, I think we can always shade the Jackson in the runtime jar instead of produce a new variant.

ShimonSte: We were planning to have two variants one shaded (for databricks) and one not (I am not strongly apposed to have only one shaded version, our rational was not to force users with a shaded jar when they don't need it).
We have managed to reach out to databricks yesterday, they will check if the shading issue could be fixed for us.
I'll changed the PR to draft for several days to give them a chance to fix it.

pan3793: Class shading and relocation are quite common practices in Java projects. I lean towards having a unified runtime jar that includes the relocated Jackson classes, if it's necessary.

spark-4.0/build.gradle
```diff
@@ -44,16 +44,27 @@ project(":clickhouse-spark-runtime-${spark_binary_version}_$scala_binary_version
...
         zip64=true
         archiveClassifier=null
 
+        // Relocate Jackson classes to avoid conflicts with external versions (e.g., from Databricks Hudi)
+        // This ensures we use our bundled Jackson version and prevents class conflicts
+        relocate "com.fasterxml.jackson", "com.clickhouse.spark.shaded.jackson"
+        
+        // Exclude Jackson service files from external sources to prevent auto-discovery
+        // of incompatible modules (e.g., from Databricks Hudi)
+        exclude "META-INF/services/com.fasterxml.jackson.databind.Module"
```
pan3793: you should relocate it instead of excluding it ...

now sure if the solution provided by https://github.com/GradleUp/shadow/issues/105 still works, as gradle is famous for its breaking changes.

spark-4.0/build.gradle
```diff
@@ -44,16 +44,27 @@ project(":clickhouse-spark-runtime-${spark_binary_version}_$scala_binary_version
...
+        
         mergeServiceFiles()
```
pan3793, line 64: oh wait, it is already here, then the runtime jar should include a relocated service manifest, why is it still in conflict with hudi?

deploy.gradle
```diff
@@ -84,8 +84,8 @@ subprojects {
 
                         pom {
                             name = "Spark ClickHouse Connector"
-                            url = "https://github.com/clickhouse/spark-clickhouse-connector"
                             description = "Spark ClickHouse Connector build on Apache Spark DataSourceV2 API."
+                            url = "https://github.com/clickhouse/spark-clickhouse-connector"
```
pan3793, line 87: please revert unnecessary changes.

spark-3.3/clickhouse-spark/src/main/scala/com/clickhouse/spark/ClickHouseTableProvider.scala
```diff
@@ -0,0 +1,215 @@
...
+      .getOrElse {
+        throw new IllegalArgumentException(
+          "Cannot auto-generate ORDER BY clause: all columns are nullable. " +
+            "ClickHouse Cloud requires non-nullable columns in ORDER BY. " +
+            "Please either: (1) make at least one column non-nullable, or " +
+            "(2) explicitly set 'order_by' option to specify which column(s) to use."
+        )
+      }
+
+    log.info(s"Auto-generated ORDER BY clause: $orderByColumn")
```
pan3793: I'm not sure if such an automatic behavior is a good idea, could you please elaborate more about your consideration? and this at least requires a config to allow the user to disable it.

ShimonSte: The consideration was not to force users who don't care about the order by to insert one and use the first Non null column.
In hindsight this is not a good idea as clickhouse itself forces the order by.
I'll change the code so that it will throw an error if the orderby clause is missing

spark-4.0/build.gradle
ShimonSte: The relocate was introduced in this PR.
The exclude was a extra "Safety measure", i checked and it is really not needed.
Removing it

spark-4.0/clickhouse-spark/src/main/scala/com/clickhouse/spark/ClickHouseTableProvider.scala
```diff
@@ -0,0 +1,222 @@
...
+
+    val optionsMap = options.asCaseSensitiveMap().asScala.toMap
+    val normalizedOptions = optionsMap ++ Map(
+      Constants.CATALOG_PROP_HOST -> optionsMap.getOrElse("host", ""),
+      Constants.CATALOG_PROP_DATABASE -> optionsMap.getOrElse("database", "default"),
+      Constants.CATALOG_PROP_HTTP_PORT -> optionsMap.getOrElse("http_port", "8123"),
+      Constants.CATALOG_PROP_PROTOCOL -> optionsMap.getOrElse("protocol", "http"),
+      Constants.CATALOG_PROP_USER -> optionsMap.getOrElse("user", "default"),
+      Constants.CATALOG_PROP_PASSWORD -> optionsMap.getOrElse("password", "")
+    ) ++ (
```
mzitnik, line 206: @ShimonSte looking at `def createCatalog` at spark 3.3/3.4/3.5 implementations, it's missing this section
```
if (optionsMap.contains("timezone")) {
        Map(Constants.CATALOG_PROP_TZ -> optionsMap("timezone"))
      } else {
        Map.empty[String, String]
      }
```

spark-4.0/clickhouse-spark/src/main/scala/com/clickhouse/spark/ClickHouseTableProvider.scala
```diff
@@ -0,0 +1,222 @@
...
+      if (optionsMap.contains("timezone")) {
```
mzitnik: Here we can replace `CATALOG_PROP_TZ` instead `timezone`

spark-4.0/clickhouse-spark/src/main/scala/com/clickhouse/spark/ClickHouseTableProvider.scala
ShimonSte: Good catch

mzitnik:
@ShimonSte, can you confirm that all our versions (Spark 3.3/3.4/3.5/4.0) are aligned with the same code

mzitnik (approved):

Issue #475, opened 2025-12-10
Problem loading UInt64 type from ClickHouse


Castiron-hub:
Incorrect Type Mapping: ClickHouse UInt64 Cannot Be Safely Converted to Java/Scala Long.

I see that my application does not work correctly when working with the Uint64 type when loading data from ClickHouse when using Spark in json/binary modes.

### Example Code:
https://github.com/ClickHouse/spark-clickhouse-connector/blob/main/spark-3.5/clickhouse-spark/src/main/scala/org/apache/spark/sql/clickhouse/SchemaUtils.scala

` case UInt32 | Int64 | UInt64 => LongType`

The current type mapping implementation incorrectly converts ClickHouse UInt64 type to Java/Scala Long type. This causes potential data loss and overflow issues because:

ClickHouse UInt64 range: 0 to 18,446,744,073,709,551,615
Java/Scala Long range: -9,223,372,036,854,775,808 to 9,223,372,036,854,775,807
The maximum value of UInt64 (≈18.4 quintillion) exceeds the maximum value of Java/Scala Long (≈9.2 quintillion)

leading to:
- Data truncation for values > 9,223,372,036,854,775,807
- Incorrect negative values due to overflow
- Potential runtime exceptions

PR #477, opened 2025-12-16
Fix UInt64 type mapping to prevent data loss and overflow

ShimonSte:
Solving issue #475

ClickHouse `UInt64` type was incorrectly mapped to Spark's `LongType`, causing data loss and overflow issues. The maximum value of `UInt64` (18,446,744,073,709,551,615) exceeds the maximum value of Java/Scala `Long` (9,223,372,036,854,775,807), leading to:
- Data truncation for values > `Long.MAX_VALUE`
- Incorrect negative values due to overflow
- Potential runtime exceptions

## Solution

Changed `UInt64` mapping from `LongType` to `DecimalType(20, 0)` in `SchemaUtils.scala`. `DecimalType(20, 0)` can safely represent the full `UInt64` range (0 to 18,446,744,073,709,551,615).

## Changes

- Updated type mapping in `SchemaUtils.scala` for Spark 3.3, 3.4, 3.5, and 4.0
- Updated tests to use `BigDecimal` for `UInt64` values to support the full range
- Updated test assertions to use `getDecimal()` instead of `getLong()` for `UInt64`

### DIFF
--- a/spark-3.3/clickhouse-spark-it/src/test/scala/org/apache/spark/sql/clickhouse/single/ClickHouseReaderTestBase.scala
+++ b/spark-3.3/clickhouse-spark-it/src/test/scala/org/apache/spark/sql/clickhouse/single/ClickHouseReaderTestBase.scala
@@ -1325,16 +1325,23 @@ trait ClickHouseReaderTestBase extends SparkClickHouseSingleTest {
           s"""INSERT INTO $actualDb.test_uint64_null VALUES
              |(1, 0),
              |(2, NULL),
-             |(3, 9223372036854775807)
+             |(3, 9223372036854775808),
+             |(4, 15000000000000000000),
+             |(5, 18446744073709551615)
              |""".stripMargin
         )
 
         val df = spark.sql(s"SELECT key, value FROM $actualDb.test_uint64_null ORDER BY key")
         val result = df.collect()
-        assert(result.length == 3)
-        assert(result(0).getDecimal(1).longValue() == 0L)
+        assert(result.length == 5)
+        assert(result(0).getDecimal(1) == BigDecimal(0))
         assert(result(1).isNullAt(1))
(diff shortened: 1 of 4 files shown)


spark-3.3/clickhouse-spark-it/src/test/scala/org/apache/spark/sql/clickhouse/single/ClickHouseDataTypeSuite.scala
```diff
@@ -94,8 +94,8 @@ abstract class ClickHouseDataTypeSuite extends SparkClickHouseSingleTest {
     ("Int32", -2147483648, 2147483647),
     ("UInt32", 0L, 4294967295L),
     ("Int64", -9223372036854775808L, 9223372036854775807L),
-    // Only overlapping value range of both the ClickHouse type and the Spark type is supported
-    ("UInt64", 0L, 4294967295L),
+    // UInt64 is mapped to DecimalType(20, 0) to support the full range: 0 to 18446744073709551615
+    ("UInt64", BigDecimal(0), BigDecimal("18446744073709551615")),
```
mzitnik, line 98: I think it will be better to use BigInteger

mzitnik (commented):
Let's use BigInteger instead BigDecimal (We do not really need to hold BigDecimal)
- memory overhead
- performnce implications

spark-3.3/clickhouse-spark-it/src/test/scala/org/apache/spark/sql/clickhouse/single/ClickHouseDataTypeSuite.scala
ShimonSte: Spark does not have a BigInteger type, thats the reason we use Sparks BigDecimal which converts to the BigDecimal you are looking at

mzitnik (approved):

Issue #382, opened 2025-01-27
Incorrect Override of useNullableQuerySchema Method in ClickHouseTable Class


jonasasx:
The useNullableQuerySchema method has been overridden in the com.clickhouse.spark.ClickHouseTable class. However, the correct place for this override is in the com.clickhouse.spark.ClickHouseCatalog class because the org.apache.spark.sql.connector.catalog.TableCatalog interface already includes this method with the following default implementation:

java
Копировать
Редактировать
default boolean useNullableQuerySchema() {
    return true;
}
Overriding the method in ClickHouseCatalog will ensure proper functionality when interacting with the catalog and avoid potential inconsistencies with the TableCatalog interface implementation.

Steps to Fix:

Move the useNullableQuerySchema method override from ClickHouseTable to ClickHouseCatalog.
Ensure the method implementation meets the requirements and returns appropriate values based on the context.
Update existing tests or add new ones to validate the method’s functionality in the ClickHouseCatalog class.
Expected Outcome: The useNullableQuerySchema method is correctly implemented in the ClickHouseCatalog class, aligning with the TableCatalog interface standards and providing the expected functionality.

Marais:
Issue was already fixed: https://github.com/ClickHouse/spark-clickhouse-connector/pull/516

PR #516, opened 2026-02-11
fix: Move useNullableQuerySchema override from ClickHouseTable to ClickHouseCatalog

veniciosAlcantara:
## Summary
Fixes [https://github.com/ClickHouse/spark-clickhouse-connector/issues/382](vscode-file://vscode-app/snap/code/219/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)

The `useNullableQuerySchema` method was incorrectly placed in `ClickHouseTable` instead of `ClickHouseCatalog`. The method was introduced to support [SPARK-43390](vscode-file://vscode-app/snap/code/219/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html), which added `useNullableQuerySchema` to the `org.apache.spark.sql.connector.catalog.TableCatalog` interface. Since `ClickHouseTable` does not implement `TableCatalog`, the override had no effect — the method was never called by the Spark engine.

This PR moves the override to `ClickHouseCatalog`, which is the actual `TableCatalog` implementation, ensuring Spark properly picks up the configured value from `spark.clickhouse.read.useNullableQuerySchema`.

## Changes

**Spark 3.5 and Spark 4.0** (where `TableCatalog.useNullableQuerySchema` exists):

**`ClickHouseCatalog.scala`**: Added `SQLConfHelper` mixin and `override def useNullableQuerySchem`a that reads from `USE_NULLABLE_QUERY_SCHEMA` configuration.
**`ClickHouseTable.scala`**: Removed the dead `useNullableQuerySchema` method and unused `USE_NULLABLE_QUERY_SCHEMA` import.

**Spark 3.3 and Spark 3.4** (where `TableCatalog` does not have this method):

**`ClickHouseTable.scala`**: Removed the dead `useNullableQuerySchema` method and unused `USE_NULLABLE_QUERY_SCHEMA` import, since `TableCatalog` in these Spark versions does not define this method.

## Checklist

- [x] Unit and integration tests covering the common scenarios were added
- [ ] A human-readable description of the changes was provided to include in CHANGELOG

---

> [!NOTE]
> **Low Risk**
> Small, targeted change to a Spark catalog override and related tests; primary risk is behavioral change in schema nullability for `CREATE/REPLACE TABLE ... AS SELECT` on Spark 3.5+/4.0 when the config is set.
> 
> **Overview**
> Fixes SPARK-43390 nullable-schema handling by **moving the `useNullableQuerySchema` override to `ClickHouseCatalog` (Spark 3.5/4.0)**, so Spark reads `spark.clickhouse.useNullableQuerySchema` from the actual `TableCatalog`.
> 
> Removes the now-dead `useNullableQuerySchema` method and related `USE_NULLABLE_QUERY_SCHEMA` imports from `ClickHouseTable` across Spark 3.3–4.0, and updates Spark 3.5/4.0 integration tests to assert nested nullability is respected when the config disables nullable query schemas.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0ecdc5136c8f43ba0d481bc468bdd021511324d1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/spark-3.3/clickhouse-spark/src/main/scala/com/clickhouse/spark/ClickHouseTable.scala
+++ b/spark-3.3/clickhouse-spark/src/main/scala/com/clickhouse/spark/ClickHouseTable.scala
@@ -23,7 +23,7 @@ import scala.collection.JavaConverters._
 import org.apache.spark.sql.catalyst.{InternalRow, SQLConfHelper}
 import org.apache.spark.sql.catalyst.expressions.GenericInternalRow
 import org.apache.spark.sql.clickhouse.{ExprUtils, ReadOptions, WriteOptions}
-import org.apache.spark.sql.clickhouse.ClickHouseSQLConf.{READ_DISTRIBUTED_CONVERT_LOCAL, USE_NULLABLE_QUERY_SCHEMA}
+import org.apache.spark.sql.clickhouse.ClickHouseSQLConf.READ_DISTRIBUTED_CONVERT_LOCAL
 import org.apache.spark.sql.connector.catalog._
 import org.apache.spark.sql.connector.catalog.TableCapability._
 import org.apache.spark.sql.connector.expressions.Transform
@@ -95,9 +95,6 @@ case class ClickHouseTable(
   }
 
   override def name: String = s"${wrapBackQuote(spec.database)}.${wrapBackQuote(spec.name)}"
-
-  // for SPARK-43390
-  def useNullableQuerySchema: Boolean = conf.getConf(USE_NULLABLE_QUERY_SCHEMA)
 
   override def capabilities(): util.Set[TableCapability] =
     Set(
--- a/spark-3.4/clickhouse-spark/src/main/scala/com/clickhouse/spark/ClickHouseTable.scala
+++ b/spark-3.4/clickhouse-spark/src/main/scala/com/clickhouse/spark/ClickHouseTable.scala
@@ -16,7 +16,7 @@ package com.clickhouse.spark
 
 import org.apache.spark.sql.catalyst.expressions.GenericInternalRow
 import org.apache.spark.sql.catalyst.{InternalRow, SQLConfHelper}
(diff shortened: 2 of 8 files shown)


ShimonSte:
Thanks for the fix @veniciosAlcantara! 
Let's align the tests in `ClickHouseDataTypeSuite` with the changes here.

veniciosAlcantara:
@ShimonSte already done.

ShimonSte:
Hi @veniciosAlcantara,
Sorry for the delay, We had found a bug in the main repo which caused some tests not to run.
Could you rebase on top of the Main branch and rerun the tests, there seems to be issues

veniciosAlcantara:
@ShimonSte, rebased on top of main (picked up #519 and #518).

After rebasing, the `ClickHouseSingleDataTypeSuite` "write supported data types" test was failing in Spark 3.5 and 4.0. The root cause:

- The test creates a table via Spark SQL DDL (`CREATE TABLE ... USING ClickHouse`), which produces `ARRAY<STRING>` and `MAP<STRING, STRING>` in the DDL.
- Spark's DDL parser does not preserve `containsNull`/`valueContainsNull` flags. They default to `true` when parsed back. This causes the ClickHouse table to be created with `Array(Nullable(String))` instead of `Array(String)`.
- When reading back, the schema correctly reflects what's stored: `ArrayType(StringType, true).` But the test asserted against the original schema which had `containsNull = false`.

This is a Spark SQL DDL limitation, not a bug in the useNullableQuerySchema fix. Our fix works correctly, top-level field nullability (`StructField.nullable = false`) is properly preserved by SPARK-43390. We verified this by checking the ClickHouse Java client (v0.9.5) which correctly reports `isNullable = false` for nested `String` inside `Array(String)`.

Fix: Updated the test assertion in both Spark 3.5 and 4.0 `ClickHouseDataTypeSuite` to normalize `containsNull`/`valueContainsNull` to `true` in the expected schema, with a comment explaining the DDL limitation. All 18 tests pass for both versions.

No production code was changed, only the test assertion logic.

image

ShimonSte (approved):

PR #536, opened 2026-04-20
fix: remove false-positive Databricks signatures from RuntimeDetector

ShimonSte:
`com.databricks.spark.util.DatabricksLogging` is present in the open-source `delta-spark` library (e.g. delta-spark_2.12:3.0.0), which is bundled by AWS Glue 5.0. This caused Glue 5 environments to be misidentified as Databricks in detectViaClassLoader().

Changes:
- Move Databricks to last in the platform check order so more class-exclusive platforms (Glue, EMR, Dataproc) are matched first
- Remove DatabricksLogging and DBUtilsHolder from Databricks signatures (both present in public Maven artifacts)
- Remove leaked class names from EMR, Dataproc, HDInsight signatures
- Extract defaultClassExists as a named private method so the function can be injected in tests

Classes intentionally excluded (exist in public artifacts):
- com.databricks.spark.util.DatabricksLogging    (io.delta:delta-spark)
- com.databricks.dbutils_v1.DBUtilsHolder         (com.databricks:dbutils-api_2.12)
- com.amazon.emr.kinesis.client.KinesisConnector  (com.amazonaws:amazon-kinesis-connectors)
- com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystem (com.google.cloud.bigdataoss:gcs-connector)
- com.google.cloud.spark.bigquery.BigQueryConnector     (com.google.cloud.spark:spark-bigquery-with-dependencies)
- com.microsoft.hdinsight.spark.common.HttpFutureCallback (azure-mgmt-hdinsight transitive dep)

Also reorders detectViaStackTrace() to check Databricks last for the same reason: databricks.spark.* patterns appear in delta-spark stack frames on non-Databricks platforms.

### DIFF
--- a/clickhouse-core/src/main/scala/com/clickhouse/spark/Utils.scala
+++ b/clickhouse-core/src/main/scala/com/clickhouse/spark/Utils.scala
@@ -196,7 +196,10 @@ object Utils extends Logging {
         .orElse(RuntimeDetector.detectViaThreadNames())
 
     /**
-     * Examines the current stack trace and loaded classes for platform-specific signatures
+     * Examines the current stack trace and loaded classes for platform-specific signatures.
+     * Platforms are checked from most class-exclusive to least, with Databricks last because
+     * some of its class patterns (e.g. databricks.spark.*) leak into open-source libraries
+     * such as delta-spark and can produce false positives on other platforms.
      */
     def detectViaStackTrace(): Option[String] = {
       val stackTrace = Thread.currentThread().getStackTrace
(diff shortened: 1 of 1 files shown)


BentsiLeviav (approved):
@ShimonSte, before we merge this, could you confirm it was verified in Glue & Databricks?
I know you use EMR in the LT. Could you verify that as well?

ShimonSte:
@BentsiLeviav It was confirmed in GLUE, DB, EMR

PR #538, opened 2026-04-25
Preventing extra quotes in filter expr when cols have special characters

### LINKED ISSUE #534: Double or unnecessary Quoting of Columns in the filters
There appears to be an inconsistency in how column quoting is handled.

In Spark SQL, column names are already quoted when required (e.g., columns containing spaces or special characters) before they reach the clickhouse connector layer.

However, in the connector’s `SQLHelper.compileFilter` method, column names are always quoted unconditionally, regardless of whether they were already quoted by Spark.

This results in double quoting (e.g., space col → \``space col\`` )

**Suggested Fix** 
```
def quoted(token: String): String =
    if (token.startsWith("`") && token.endsWith("`")) token
    else s"`$token`"
```
**Schema of table**
```
1. │ CREATE TABLE default.space_col   ↴│
   │↳(                                ↴│
   │↳    `space col` Nullable(String),↴│
   │↳    `col` Int32 DEFAULT 1        ↴│
   │↳)                                ↴│
   │↳ENGINE = MergeTree               ↴│
   │↳ORDER BY tuple()                 ↴│
   │↳SETTINGS index_granularity = 8192 │
```
**Piece of code to Reproduce the issue:**
```
Dataset<Row> ds = spark.read()
            .format("clickhouse")
            ...
            .load();

ds.filter(functions.col("col").isNotNull()) .show(2, false); // Works Fine

ds.filter(functions.col("space col").isNotNull()).show(2, false);  // Fails when column contains space
```

**pom.xml**
```
<dependencies>
        <dependency>
            <groupId>com.clickhouse.spark</groupId>
            <artifactId>clickhouse-spark-runtime-3.5_2.12</artifactId>
            <version>0.10.0</version>
        </dependency>
        <dependency>
            <groupId>org.apache.spark</groupId>
            <artifactId>spark-core_2.12</artifactId>
            <version>3.5.0</version>
            <scope>compile</scope>
        </dependency>
        <dependency>
            <groupId>org.apache.spark</groupId>
            <artifactId>spark-sql_2.12</artifactId>
            <version>3.5.0</version>
            <scope>compile</scope>
        </dependency>
        <dependency>
            <groupId>com.fasterxml.jackson.core</groupId>
            <artifactId>jackson-core</artifactId>
            <version>2.15.2</version>
        </dependency>
    </dependencies>
```

**Stacktrace**
```
Caused by: com.clickhouse.client.api.ServerException: Code: 62. DB::Exception: Syntax error: failed at position 274 ('``') (line 3, col 29): ``space col`` IS NOT NULL) AND (1=1))  LIMIT 3  . Expected one of: literal, NULL, NULL, number, Bool, TRUE, FALSE, string literal, SELECT query, possibly with UNION, list of union elements, SELECT query, subquery, possibly with UNION, SELECT subquery, SELECT query, WITH, FROM, SELECT, EXPLAIN, EXPLAIN, token, Comma, ClosingRoundBracket, CAST operator, NOT, INTERVAL, CASE, DATE, TIMESTAMP, tuple, collection of literals, array, asterisk, qualified asterisk, compound identifier, identifier, COLUMNS matcher, COLUMNS, qualified COLUMNS matcher, function name, substitution, MySQL-style global variable. (SYNTAX_ERROR) (version 25.1.1.4165 (official build))"] 
	at com.clickhouse.client.api.internal.HttpAPIClientHelper.readError(HttpAPIClientHelper.java:403)
	at com.clickhouse.client.api.internal.HttpAPIClientHelper.executeRequest(HttpAPIClientHelper.java:467)
	at com.clickhouse.client.api.Client.lambda$query$5(Client.java:1599)
	at com.clickhouse.client.api.Client.runAsyncOperation(Client.java:2012)
	at com.clickhouse.client.api.Client.query(Client.java:1644)
	at com.clickhouse.client.api.Client.query(Client.java:1542)
	at com.clickhouse.spark.client.NodeClient.$anonfun$queryAndCheck$2(NodeClient.scala:236)
	at scala.util.Try$.apply(Try.scala:210)
	at com.clickhouse.spark.client.NodeClient.queryAndCheck(NodeClient.scala:236)
	... 36 more

```
**Debugging Notes**
1. Verified via debugging `PushDownUtils.pushFilters` (Spark SQL) that:
1.1 Columns are already quoted only when necessary
1.2 Example: "col" → col  and "space col" → \`space col\`
2. By the time filters reach the connector, quoting has already been correctly applied
3. The issue arises when `SQLHelper.compileFilter` applies quoting again

shashank-gb:
## Summary
Fixes #534: Prevent double quoting of column names in `SQLHelper`

Spark already quotes column names when needed, but the connector was quoting them again, causing invalid SQL for columns with spaces (e.g. \``space col\``). This fix ensures column names are only quoted if not already quoted.

## Checklist
Delete items not relevant to your PR:
- [x] Unit and integration tests covering the common scenarios were added
- [x] A human-readable description of the changes was provided to include in CHANGELOG

### DIFF
--- a/spark-3.3/clickhouse-spark/src/main/scala/com/clickhouse/spark/SQLHelper.scala
+++ b/spark-3.3/clickhouse-spark/src/main/scala/com/clickhouse/spark/SQLHelper.scala
@@ -16,6 +16,7 @@ package com.clickhouse.spark
 
 import java.sql.{Date, Timestamp}
 import java.time.{Instant, LocalDate, LocalDateTime, ZoneId}
+import java.util.regex.Pattern
 import org.apache.commons.lang3.StringUtils
 import org.apache.spark.sql.connector.expressions.aggregate._
 import org.apache.spark.sql.connector.expressions.NamedReference
@@ -25,7 +26,13 @@ import com.clickhouse.spark.Utils._
 
 trait SQLHelper {
 
-  def quoted(token: String) = s"`$token`"
+  private val validIdentPattern = Pattern.compile("^[a-zA-Z_][a-zA-Z0-9_]*")
+
+  private def alreadyQuoted(part: String): Boolean =
+    !validIdentPattern.matcher(part).matches()
+
+  def quoted(token: String): String =
+    if (alreadyQuoted(token)) token else s"`$token`"
 
   // null => null, ' => ''
   def escapeSql(value: String): String = StringUtils.replace(value, "'", "''")
--- a/spark-3.3/clickhouse-spark/src/main/scala/com/clickhouse/spark/SQLHelper.scala
+++ b/spark-3.3/clickhouse-spark/src/main/scala/com/clickhouse/spark/SQLHelper.scala
@@ -26,13 +26,10 @@ import com.clickhouse.spark.Utils._
 
 trait SQLHelper {
 
-  private val validIdentPattern = Pattern.compile("^[a-zA-Z_][a-zA-Z0-9_]*")
-
-  private def alreadyQuoted(part: String): Boolean =
-    !validIdentPattern.matcher(part).matches()
-
-  def quoted(token: String): String =
-    if (alreadyQuoted(token)) token else s"`$token`"
+  def quoted(token: String): String = token match {
+    case t if t.startsWith("`") && t.endsWith("`") => t
+    case t => s"`$t`"
+  }
 
   // null => null, ' => ''
   def escapeSql(value: String): String = StringUtils.replace(value, "'", "''")
--- a/spark-3.3/clickhouse-spark/src/main/scala/com/clickhouse/spark/SQLHelper.scala
+++ b/spark-3.3/clickhouse-spark/src/main/scala/com/clickhouse/spark/SQLHelper.scala
@@ -16,7 +16,6 @@ package com.clickhouse.spark
 
 import java.sql.{Date, Timestamp}
 import java.time.{Instant, LocalDate, LocalDateTime, ZoneId}
-import java.util.regex.Pattern
 import org.apache.commons.lang3.StringUtils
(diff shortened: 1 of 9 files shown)


spark-3.3/clickhouse-spark/src/main/scala/com/clickhouse/spark/SQLHelper.scala
ShimonSte, line 35: This function will return True for any non-simple-identifier,
I think a better approach would be to try a pattern match vs a regex pattern

```suggestion
   def quoted(token: String): String = token match {
    case t if t.startsWith("`") && t.endsWith("`") => t
    case t                                         => s"`$t`"
  }
```

shashank-gb: Any other suggestions, is test case is in the correct place ?

spark-3.4/clickhouse-spark/src/main/scala/com/clickhouse/spark/SQLHelper.scala
```diff
@@ -16,6 +16,7 @@ package com.clickhouse.spark
 
 import java.sql.{Date, Timestamp}
 import java.time.{Instant, LocalDate, LocalDateTime, ZoneId}
+import java.util.regex.Pattern
```
ShimonSte: Could you also remove the import from the rest of the packages

shashank-gb: done!

ShimonSte (approved):
Nice job @shashank-gb, thank you for contributing

shashank-gb:
Woo hoo, Thanks @ShimonSte

PR #539, opened 2026-05-04
Test/withtable extra properties

ShimonSte:
This PR comes to align all tests.
In most tests we use the withTable helper function. some tests were not using it.
This PR introduced a small change to allow customizing the tableProperties in the withTable method hence allowing all tests to be converted to the withTable function

### DIFF
--- a/spark-3.3/clickhouse-spark-it/src/test/scala/org/apache/spark/sql/clickhouse/single/ClickHouseWriterTestBase.scala
+++ b/spark-3.3/clickhouse-spark-it/src/test/scala/org/apache/spark/sql/clickhouse/single/ClickHouseWriterTestBase.scala
@@ -974,11 +974,6 @@ trait ClickHouseWriterTestBase extends SparkClickHouseSingleTest {
   }
 
   test("write StructType - full round-trip with Spark-created table") {
-    // This test verifies the complete cycle: Spark creates table, writes data, reads it back
-    val db = "test_db"
-    val tbl = "test_struct_roundtrip"
-
-    // Create schema with various struct configurations
     val schema = StructType(
       StructField("id", LongType, false) ::
         StructField(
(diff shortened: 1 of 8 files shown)


spark-3.3/clickhouse-spark-it/src/test/scala/org/apache/spark/sql/clickhouse/single/ClickHouseWriterTestBase.scala
```diff
@@ -1001,24 +996,7 @@ trait ClickHouseWriterTestBase extends SparkClickHouseSingleTest {
...
-           |  metadata STRUCT<created: DATE, score: DOUBLE, tags: ARRAY<STRING>> NOT NULL
-           |) USING clickhouse
-           |TBLPROPERTIES (
-           |  engine = 'MergeTree()',
-           |  order_by = 'id'
-           |)
-           |""".stripMargin
-      )
-      // Write data via Spark
+    withTable("test_db", "test_struct_roundtrip", schema) { (actualDb, actualTbl) =>
```
mzitnik, line 999: I do not see any hints for 
> |TBLPROPERTIES (
> |  engine = 'MergeTree()',
> |  order_by = 'id'
> |)

Is this hardcoded in the schema?

mzitnik (commented):
Maybe we can extract the withTable to some shared package

spark-3.3/clickhouse-spark-it/src/test/scala/org/apache/spark/sql/clickhouse/single/ClickHouseWriterTestBase.scala
ShimonSte: Its hardcoded as the default values of the  `withTable` method

ShimonSte:
> Maybe we can extract the withTable to some shared package

This was a conversation we had once with Pan, He was very against creating shared code in this situation.
He took this pattern from projects like Apache Iceberg where they have different code for each Spark version.
The rational behind it is to have the flexibility to adapt to Spark's new versions without breaking old versions.
I can look for the conversation if you are interested.

mzitnik (approved):

PR #541, opened 2026-05-05
Fix/issue 521 arrow nested variant

ShimonSte:
Writing a Spark `VariantType` column to a ClickHouse `Array(JSON)` column via Arrow format failed with `CANNOT_PARSE_QUOTED_STRING`. The Arrow writer's existing workaround (rewrite `VariantType → StringType` and JSON-stringify each `VariantVal` per row) only inspected top-level fields, so any `VariantType` nested inside `Array`, `Map`, or `Struct` was passed through unchanged and ClickHouse rejected the input.                                
                                                                                                                                                         
  ## Changes                                                                                                                                             
                                                                                                                                                         
  - New `VariantUtils` object that recursively rewrites a schema (`VariantType → StringType` at any depth) and rewrites a row's values (every `VariantVal` → JSON `UTF8String`).                                                                                                                   
  - `ClickHouseArrowStreamWriter` slimmed down to delegate to `VariantUtils` — the previous top-level-only logic is gone.                                
  - Four new integration tests in `ClickHouseWriterTestBase` covering `Array<VARIANT>` (the issue repro), `Map<STRING, VARIANT>`, `STRUCT` containing `VARIANT`, and `Array<STRUCT>` containing `VARIANT`. They run under both the JSON and Arrow writer suites.

### DIFF
--- a/spark-4.0/clickhouse-spark/src/main/scala/org/apache/spark/sql/clickhouse/VariantUtils.scala
+++ b/spark-4.0/clickhouse-spark/src/main/scala/org/apache/spark/sql/clickhouse/VariantUtils.scala
@@ -15,11 +15,11 @@
 package org.apache.spark.sql.clickhouse
 
 import org.apache.spark.sql.catalyst.InternalRow
-import org.apache.spark.sql.catalyst.expressions.SpecializedGetters
+import org.apache.spark.sql.catalyst.expressions.{GenericInternalRow, SpecializedGetters}
 import org.apache.spark.sql.catalyst.util.{ArrayBasedMapData, ArrayData, GenericArrayData, MapData}
 import org.apache.spark.sql.types.{ArrayType, DataType, MapType, StringType, StructType, VariantType}
+import org.apache.spark.types.variant.Variant
 import org.apache.spark.unsafe.types.{UTF8String, VariantVal}
-
 
 object VariantUtils {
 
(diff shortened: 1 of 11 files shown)


spark-4.0/clickhouse-spark/src/main/scala/org/apache/spark/sql/clickhouse/VariantUtils.scala
```diff
@@ -0,0 +1,90 @@
...
+    val variant = new Variant(variantVal.getValue, variantVal.getMetadata)
+    UTF8String.fromString(variant.toJson(java.time.ZoneId.of("UTC")))
+  }
+
+  /**
+   * Recursively rewrite `value`, converting every `VariantVal` at any depth into a JSON
+   * UTF8String. `dt` is the *original* (pre-rewrite) Spark type for `value`. Subtrees that
+   * contain no Variant are returned as-is, so the cost is paid only along Variant-bearing paths.
+   */
+  def convertVariantsToJson(value: Any, dt: DataType): Any = (value, dt) match {
```
mzitnik, line 63: As we discussed, the programmatic recursion can be risky in some situations.
One alternative, instead of implementing an iterative we can place a depth guard inside the method

mzitnik (approved):
Approving, but please see my comments and also co-pilot review

mzitnik (approved):
LGTM

PR #543, opened 2026-05-10
feat: implement SupportsPushDownTopN (top-N pushdown) across spark-3.3/3.4/3.5/4.0

ShimonSte:
Push down `ORDER BY ... LIMIT n` (top-N) to ClickHouse for the DataSource V2 scan across all four Spark profiles (3.3, 3.4, 3.5, 4.0). Each input partition issues its own local top-N and Spark performs a final global merge across partitions (`isPartiallyPushed = true`). Gated by a new SQL conf `spark.clickhouse.read.pushdown.topN` (default `true`).

Sort orders are translated via `ExprUtils.toClickHouseSortOrderOpt`:
- **spark-3.4 / 3.5 / 4.0**: accepts top-level `NamedReference` and `ApplyTransform` whose name is in the `FunctionRegistry`. Args are recursively walked -- column refs become `FieldRef` (back-quoted by the renderer), nested transforms recurse, literals pass through. Anything else (multi-part refs, arbitrary scalar expressions, unregistered functions) yields `None` and the top-N is left for Spark to evaluate.
- **spark-3.3**: restricted to top-level `NamedReference` and `IdentityTransform(FieldReference(Seq(_)))` only. 3.3 has no `FunctionRegistry` to gate `ApplyTransform` against; forwarding arbitrary Spark function names to ClickHouse without translation would produce SQL the server rejects at execution time, so the path is deliberately narrower than 3.4+.

Identifier back-quoting in the pushed-down clause is handled by a recursive `ClickHouseScanBuilder.renderClickHouseSql` walker so reserved-word column names round-trip safely whether at the top level (`ORDER BY \`interval\``) or nested inside a function transform (`xxHash64(\`interval\`)`). The DDL translator `ExprUtils.toClickHouse` (used by `PARTITION BY`) is intentionally unchanged -- DDL is user-typed and the user owns identifier quoting there.

Verified against a containerized ClickHouse 25.3: top-N reaches the server as the expected `ORDER BY <expr> ASC|DESC NULLS FIRST|LAST LIMIT n` clause. Aggregation pushdown is unaffected -- Spark's V2 rule does not invoke `pushTopN` over a partially-pushed Aggregate (the `PhysicalOperation` extractor doesn't traverse `Aggregate` and the connector's `SupportsPushDownAggregates.supportCompletePushDown` returns `false`), so the two pushdowns stay independent.

Includes an end-to-end integration test (`ClickHouseGenericSuite."push down top N"`) that asserts both the result rows and the wire SQL via `system.query_log`, using `interval` (CH-reserved, Spark-`ansiNonReserved`) as the sort column to genuinely exercise the back-quoting regression case.

---

> [!NOTE]
> **Medium Risk**
> Changes read query generation and partial pushdown semantics; incorrect sort translation or merge assumptions could return wrong top-N results, though unsupported sorts are rejected and the feature is disableable via config.
> 
> **Overview**
> Adds **top-N pushdown** for ClickHouse reads: eligible `ORDER BY ... LIMIT n` plans are translated into ClickHouse SQL on each scan partition, with **`isPartiallyPushed = true`** so Spark still does a global merge. Controlled by **`spark.clickhouse.read.pushdown.topN`** (default `true`), documented in SQL config docs.
> 
> **`ClickHouseScanBuilder`** implements `SupportsPushDownTopN`; sort keys are converted via **`ExprUtils.toClickHouseSortOrderOpt`** and rendered with new shared **`ExprRender`** (back-quotes identifiers such as reserved-word column `interval`). **`ScanJobDescription` / `ClickHouseReader`** carry an **`orderByClause`** into the generated `SELECT`.
> 
> **Spark 3.3** only pushes bare columns / `IdentityTransform`; **3.4+** also allow **`FunctionRegistry`**-mapped function transforms (with recursive args). Unit tests (`ExprUtilsSuite`) and integration **`push down top N`** (result + `system.query_log`) cover quoting and the conf-off path.
> 
> **`ClickHouseGenericSuite`** gains Cloud read retries (`eventuallyOnCloud`), safer parallel CI naming, UInt64 JSON setting pin, and assorted fixes using resolved `actualDb`/`actualTbl`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c56c7af2f8ac83a42f1e22e53246478c2ff0465a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/spark-4.0/clickhouse-spark-it/src/test/scala/org/apache/spark/sql/clickhouse/single/ClickHouseGenericSuite.scala
+++ b/spark-4.0/clickhouse-spark-it/src/test/scala/org/apache/spark/sql/clickhouse/single/ClickHouseGenericSuite.scala
@@ -43,12 +43,15 @@ abstract class ClickHouseGenericSuite extends SparkClickHouseSingleTest {
   }
 
   test("clickhouse catalog") {
-    withDatabase("db_t1", "db_t2") {
-      spark.sql("CREATE DATABASE db_t1")
-      spark.sql("CREATE DATABASE db_t2")
+    val prefix = if (useSuiteLevelDatabase) s"${testDatabaseName}_cat" else "db_t"
+    val db1 = s"${prefix}1"
+    val db2 = s"${prefix}2"
+    withDatabase(db1, db2) {
+      spark.sql(s"CREATE DATABASE $db1")
+      spark.sql(s"CREATE DATABASE $db2")
       checkAnswer(
-        spark.sql("SHOW DATABASES LIKE 'db_t*'"),
-        Row("db_t1") :: Row("db_t2") :: Nil
+        spark.sql(s"SHOW DATABASES LIKE '${prefix}*'"),
+        Row(db1) :: Row(db2) :: Nil
       )
       spark.sql("USE system")
       checkAnswer(
@@ -431,9 +434,15 @@ abstract class ClickHouseGenericSuite extends SparkClickHouseSingleTest {
         )
       }
 
-      runClickHouseSQL("SYSTEM FLUSH LOGS").collect()
+      runClickHouseSQL(
+        if (isCloud) "SYSTEM FLUSH LOGS ON CLUSTER 'default'"
+        else "SYSTEM FLUSH LOGS"
+      ).collect()
+      val queryLogRef =
+        if (isCloud) "clusterAllReplicas('default', system, query_log)"
+        else "system.query_log"
       val recentQueries = runClickHouseSQL(
-        s"""SELECT query FROM system.query_log
+        s"""SELECT query FROM $queryLogRef
            |WHERE type = 'QueryFinish'
            |  AND log_comment = '$topNLogTag'
            |ORDER BY event_time DESC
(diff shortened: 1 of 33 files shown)


ShimonSte:
cursorbot

spark-4.0/clickhouse-spark-it/src/test/scala/org/apache/spark/sql/clickhouse/single/ClickHouseGenericSuite.scala
```diff
@@ -390,6 +390,65 @@ abstract class ClickHouseGenericSuite extends SparkClickHouseSingleTest {
...
+           |  (1, 10),
+           |  (2, 20),
+           |  (3, 30),
+           |  (4, 40),
+           |  (5, NULL)
+           |""".stripMargin
+      ).collect()
+
+      checkAnswer(
+        spark.sql(s"SELECT interval FROM $actualDb.$actualTbl ORDER BY interval DESC NULLS LAST LIMIT 2"),
```
mzitnik: Can we add a unique comment (aka UUID) to SQL so we can check if it exists later in the query log

mzitnik (approved):

PR #544, opened 2026-05-12
Trigger clickhouse-glue tests on every merge to main

ShimonSte:
After every merge to `main`, this workflow dispatches `spark-pr-merged` to `clickhouse-glue` with the merged SHA and waits for the resulting `glue-tests.yml` and `glue-tests-cloud.yml`
  runs. The spark job fails if either glue workflow fails.
  ## How it works

  1. Push to `main` triggers the workflow
  2. Generates a short-lived token via the shared GitHub App (scoped to `clickhouse-glue` only)
  3. Dispatches `spark-pr-merged` with `client_payload.sha = <merged commit>`; glue's test workflows override their `connector-source` submodule to that SHA, so tests run against the just-merged code
  4. Polls for the new glue runs (~2.5 min budget), then waits on each with `gh run watch --exit-status`

### DIFF
--- a/.github/workflows/trigger-glue-tests.yml
+++ b/.github/workflows/trigger-glue-tests.yml
@@ -1,0 +1,108 @@
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


mzitnik:
@ShimonSte can you please add PR description

mzitnik (approved):

PR #562, opened 2026-07-12
fix(write): sort convertLocal writes by shard number instead of sharding key

### LINKED ISSUE #560: convertLocal write flushes on every shard-key change because sort order ignores shard number, producing tiny batches
When writing to a Distributed table with spark.clickhouse.write.distributed.convertLocal=true, the writer force-flushes its buffer every time the computed shard number of the incoming row differs from the buffered rows:

// ClickHouseWriter.write
val shardNum = calcShard(record)
flush(force = shardNum != currentShardNum && currentBufferedRows > 0, currentShardNum)

The connector's required ordering (RequiresDistributionAndOrdering → ExprUtils.toSparkSortOrders) sorts each Spark partition by the sharding key value, not by the shard number derived from it (ShardUtils.calcShard = weighted unsigned modulo of the key). Consecutive sorted key runs therefore alternate shards — e.g. with 2 shards and shard_by = 'k': k=3 → shard 2, k=4 → shard 1, k=5 → shard 2, … — and nearly every distinct-key boundary triggers a forced flush.

Impact

Batches are cut far below spark.clickhouse.write.batchSize, so a task performs up to one INSERT per distinct sharding-key value instead of one per batchSize rows. This means many small HTTP inserts (per-request overhead, retries amplified) and many small MergeTree parts on the server. The double buffering also can't help: currentShardNum tracks only one shard at a time.

Repro

- 2-shard cluster, Distributed table with an integer sharding key (e.g. shard_by = 'k'), convertLocal=true, batchSize=100000.
- Write a DataFrame with many distinct k values (e.g. NYC yellow taxi, shard_by = 'pu_location_id', 260 distinct values, ~3M rows).
- Observe flush-count / batch-size metrics: batch sizes track rows-per-distinct-key, not batchSize. With unsorted input it degrades further (flush every few rows)

Example from yellow taxi dataset:

Image

Marais:
Fixes ClickHouse/spark-clickhouse-connector#560.

With `convertLocal=true` the writer flushes whenever the incoming row's shard differs from the buffered rows, but rows were sorted by sharding-key **value** — consecutive values map to different shards, so nearly every distinct key boundary forced a tiny INSERT.

**What changed**

* convertLocal writes now sort by **shard number**, via a new `clickhouse_shard_num` catalog function (Spark 3.4/3.5/4.0). `format("clickhouse")` and Spark 3.3 cannot resolve catalog functions in write sort orders and keep the previous key-value sort — gated by `functionCatalogUsable`, logged as a warning, covered by tests.
* Shard routing tables are precomputed per `ClusterSpec`; `ShardUtils.calcShard` is now allocation-free. It runs once per row in the writer and \~twice per row comparison in the sort — without this the shard-number sort regresses the write to 10.9s.

**Measured** — NYC yellow taxi, 2,964,624 rows, 4 shards, `batchSize=100000`, `shard_by='pu_location_id'` (260 distinct values). Wall times best of 3 from the compare benchmark; batch stats from `system.query_log`:

|  | write batches from Spark | avg rows/batch | write wall time |
| -- | -- | -- | -- |
| convertLocal=true, key-value sort (main today) | 226 | 13,118 | 2,807 ms |
| convertLocal=true, shard-num sort (**this PR**) | **49** | **60,503** | **2,154 ms** |
| convertLocal=false (Distributed engine routing) | 33 | 89,837 | 1,962 ms |

convertLocal=false is untouched by this PR; measured with `distributed_foreground_insert=1` so the write is synchronous and comparable. Note its batches go to the Distributed table, which splits each per shard on the server — the local tables end up receiving \~4x smaller batches (median ≈23K rows), while convertLocal=true batches land whole.

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)


BentsiLeviav:
@Marais could you verify the reason for the tests' failure?

spark-3.4/clickhouse-spark/src/main/scala/org/apache/spark/sql/clickhouse/ExprUtils.scala
```diff
@@ -57,19 +57,32 @@ object ExprUtils extends SQLConfHelper with Serializable with Logging {
...
+      .map(Expressions.sort(_, SortDirection.ASCENDING))
+    val sortingKeySortOrders = sortingKey.seq.flatten.flatten { case OrderExpr(expr, asc, nullFirst) =>
+      val direction = if (asc) SortDirection.ASCENDING else SortDirection.DESCENDING
+      val nullOrder = if (nullFirst) NullOrdering.NULLS_FIRST else NullOrdering.NULLS_LAST
+      toSparkTransformOpt(expr, functionRegistry).map(trans =>
+        Expressions.sort(trans, direction, nullOrder)
+      )
+    }
+    (shardingSortOrders ++ partitionSortOrders ++ sortingKeySortOrders).toArray
+  }
```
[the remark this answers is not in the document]
Marais: Not a regression — this misreads `++:`, which prepends its **left** operand to the right collection: `List(1, 2) ++: List(3, 4) == List(1, 2, 3, 4)`. So the old `splits.map(sort) ++: sortingKeyOrders` already evaluated to (sharding key, partition key, ORDER BY keys) — exactly the order this PR now emits explicitly. The sharding key has always been the leading required sort for Distributed writes with a non-rand sharding key, convertLocal on or off; `WriteJobDescriptionSuite` pins the ordering for all flag combinations.

Marais:
> @Marais could you verify the reason for the tests' failure?

Wasn't ready for review yet. :) Still tweaking this.

BentsiLeviav:
@Marais if that's the case, let's move this to draft

ShimonSte:
Great work!
I have Two question, 
It seems like for each row we will calculate the shard number twice (one in the `toSparkSortOrders` and one in the physical write `requiredOrdering` method inside the `ClickhouseWrite` object)
Could we do something about it? is it worth to keep the shard num as metadata?
Also WDYT about moving the whole sorting into the physical write object, then we won't calculate it twice and we also won't need to differentiate between catalog and tableProvider.

spark-4.0/clickhouse-spark-it/src/test/scala/org/apache/spark/sql/clickhouse/cluster/ClusterShardNumSortWriteSuite.scala
```diff
@@ -0,0 +1,193 @@
...
+
+      // rows arrive sorted by shard number, so each writer task issues about one INSERT per
+      // shard; without the shard-num sort it would be about one INSERT per distinct key value
+      val totalInserts = shardNodeOptions.values.map(countLocalInserts(_, db, tbl_local)).sum
+      val numShards = shardCatalogs.size
+      assert(totalInserts >= numShards, s"expected at least one INSERT per shard, got $totalInserts")
+      assert(
+        totalInserts <= numShards * numSparkPartitions + 2,
+        s"expected about one INSERT per shard per Spark partition, got $totalInserts"
+      )
```
ShimonSte, line 134: Do you think AQE can mess this up? should we turn it off?

spark-4.0/clickhouse-spark-it/src/test/scala/org/apache/spark/sql/clickhouse/cluster/ClusterShardNumSortWriteSuite.scala
```diff
@@ -0,0 +1,193 @@
...
+      "protocol" -> "http",
+      "user" -> "default",
+      "password" -> "",
+      "database" -> "default"
+    )
+  }
+
+  private def resetQueryLog(nodeOptions: Map[String, String]): Unit = {
+    runClickHouseSQL("SYSTEM FLUSH LOGS", nodeOptions)
+    runClickHouseSQL("TRUNCATE TABLE IF EXISTS system.query_log", nodeOptions)
```
ShimonSte, line 81: Don't really like what i am suggesting but lets add a sleep here to avoid flakiness

BentsiLeviav:
@Marais I see cloud tests are failing - can you check why? If it is related to replica synchronization, there are some settings that could help us overcome it.

Marais:
> Great work! I have Two question, It seems like for each row we will calculate the shard number twice (one in the `toSparkSortOrders` and one in the physical write `requiredOrdering` method inside the `ClickhouseWrite` object) Could we do something about it? is it worth to keep the shard num as metadata? Also WDYT about moving the whole sorting into the physical write object, then we won't calculate it twice and we also won't need to differentiate between catalog and tableProvider.

@ShimonSte DSV2 creates 2 operators in our case. SortExec and AppendExec. It only lets you declare the ordering as an expression or column value. There is no comms between the 2 spark operators. So not possible to share/materialize this value.

the interesting thing is how many times it will be called: 

- For the sort it will call O(nlogn) times. Interestingly, if we would sort by one column, (i.e. drop partition sort), it would do radix sort and then it wouild be O(n)
- The appendExec - calls calcShard O(n) times

So the compexity is O(nLog(n)): Important to know, I optimized the calcShard. Without the optimizatiion, because of the increased times it calls the caclShard, the perforamnce degraded significantly if we sort by expression (shardNum). 

But now, after optimizing it not to do any object allocation, the time per row to call calcShard is in the single digit ns (estimation), which should make it a low effect on the overall performance.

Marais:
> > Great work! I have Two question, It seems like for each row we will calculate the shard number twice (one in the `toSparkSortOrders` and one in the physical write `requiredOrdering` method inside the `ClickhouseWrite` object) Could we do something about it? is it worth to keep the shard num as metadata? Also WDYT about moving the whole sorting into the physical write object, then we won't calculate it twice and we also won't need to differentiate between catalog and tableProvider.
> 
> @ShimonSte DSV2 creates 2 operators in our case. SortExec and AppendExec. It only lets you declare the ordering as an expression or column value. There is no comms between the 2 spark operators. So not possible to share/materialize this value.
> 
> the interesting thing is how many times it will be called:
> 
> * For the sort it will call O(nlogn) times. Interestingly, if we would sort by one column, (i.e. drop partition sort), it would do radix sort and then it wouild be O(n)
> * The appendExec - calls calcShard O(n) times
> 
> So the compexity is O(nLog(n)): Important to know, I optimized the calcShard. Without the optimizatiion, because of the increased times it calls the caclShard, the perforamnce degraded significantly if we sort by expression (shardNum).
> 
> But now, after optimizing it not to do any object allocation, the time per row to call calcShard is in the single digit ns (estimation), which should make it a low effect on the overall performance.

Also, to move the sorting to write side: Not a good idea, means we are implemeting sort, need to handle spill etc.

spark-4.0/clickhouse-spark-it/src/test/scala/org/apache/spark/sql/clickhouse/cluster/ClusterShardNumSortWriteSuite.scala
Marais: not really, but I disabled it to be on the safe side

spark-4.0/clickhouse-spark-it/src/test/scala/org/apache/spark/sql/clickhouse/cluster/ClusterShardNumSortWriteSuite.scala
Marais: ok

ShimonSte (approved):
Great job

PR #563, opened 2026-07-14
test: reproduce SSL `ssl` option leak warning (#557)

### LINKED ISSUE #557: Connector leaks its own `ssl` option to the clickhouse-java client, causing a confusing `Unknown and unmapped config properties` warning
## What happens

When writing with SSL enabled, every job logs this warning repeatedly:

```
WARN ClientConfigProperties: Unknown and unmapped config properties: {ssl=true, ssl_mode=NONE}
```

Data is written correctly and SSL works fine - but the warning is alarming, looks like a misconfiguration, and shows up on every connection. A user recently reported it and assumed they'd set something up wrong (they hadn't).

## Repro

```python
df.write.format("clickhouse") \
    .option("host", host) \
    .option("http_port", port) \
    .option("protocol", "https") \
    .option("user", user) \
    .option("password", password) \
    .option("database", db) \
    .option("table", table) \
    .save()
```

Versions: `clickhouse-spark-runtime-4.0_2.13-0.10.0`, clickhouse-java client `0.9.5`.

## Root cause

The connector uses `option.ssl` **itself** to pick the URL scheme:

```scala
// NodeClient.scala
private def createClickHouseURL(nodeSpec: NodeSpec): String = {
  val ssl = nodeSpec.options.getOrDefault("ssl", "false").toBoolean
  if (ssl) s"https://${nodeSpec.host}:${nodeSpec.port}"
  else     s"http://${nodeSpec.host}:${nodeSpec.port}"
}
```

…and then *also* forwards the entire options map — including `ssl` — straight into the java client:

```scala
private val client = new Client.Builder()
  ...
  .setOptions(nodeSpec.options)   // ssl gets passed through here too
  .addEndpoint(createClickHouseURL(nodeSpec))
  .build()
```

The v2 java client validates option keys against its `ClientConfigProperties` enum and warns about anything it doesn't recognize (`ClientConfigProperties.parseConfigMap`). `ssl` is **not** a clickhouse-java property — it's a connector-only option — so it lands in the "unknown and unmapped" bucket and gets logged on every client build.

So `ssl` is consumed correctly by the connector **and** redundantly leaked to a client that doesn't understand it. That redundant forwarding is the bug.

(The `ssl_mode=NONE` in the same message is a separate, legacy key the user set directly — the connector doesn't read it and it's not a v2 client key either. Out of scope for this issue.)

## Why removing the option "fixes" the warning but breaks connectivity

`option.ssl` is the only thing driving `https://` vs `http://` here — `protocol=https` does not affect the HTTP client URL. Drop `option.ssl` and the endpoint falls back to `http://`, so the connection fails. Users are stuck between a noisy log and a broken connection.

## Proposed fix

Strip the options the connector consumes itself (currently `ssl`) from the map before passing it to `setOptions()`, so we don't hand connector-internal keys to a client that will reject them. No behavior change — SSL is already driven by the URL scheme.

While doing this it's worth grepping for every key the connector reads directly out of `nodeSpec.options` so the exclude list is complete rather than just `ssl`.

## Workaround for now

The warning is cosmetic and safe to ignore. To silence it, set the log level for `com.clickhouse.client.api.ClientConfigProperties` to `ERROR`.

Marais:
Fixes #557.

`NodeClient` reads `option.ssl` to pick the `https://` URL scheme but also forwarded it to the clickhouse-java client, which logged `Unknown and unmapped config properties: {ssl=true}` on every build. Now connector-consumed keys are stripped before `setOptions`. SSL is still driven by the URL scheme — no behavior change.

**Tests:** `NodeClientSuite` (offline unit) — `ssl` no longer leaks, and unknown client keys still surface. We did not include an automated SSL connectivity test because Cloud tests are flaky; instead **SSL connectivity (query + insert round-trip over https) was confirmed manually against a real cluster on staging mk-test**.

I also tested against cloud to check that we will never pass a prop to the java client if it is not supported.

image

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ShimonSte:
I think we already have a place to remove unwanted options.
Take a look at the `buildNodeSpec` method in the `ClickHouseHelper` file and the . see you can achieve this functionality by adding it to the `CATALOG_PROP_IGNORE_OPTIONS`

Marais:
> I think we already have a place to remove unwanted options. Take a look at the `buildNodeSpec` method in the `ClickHouseHelper` file and the . see you can achieve this functionality by adding it to the `CATALOG_PROP_IGNORE_OPTIONS`

@ShimonSte Made changes, but CATALOG_PROP_IGNORE_OPTIONS is not the right pace to include it. Made SSL a firt class citizen, and added failsafe check on what we pass to the client.

ShimonSte (approved):
LGTM

clickhouse-core/src/main/scala/com/clickhouse/spark/client/NodeClient.scala
```diff
@@ -87,20 +96,29 @@ class NodeClient(val nodeSpec: NodeSpec, queryTimeoutMs: Long = NodeClient.DEFAU
...
+  private def createClickHouseURL(nodeSpec: NodeSpec): String =
+    if (nodeSpec.ssl) {
       s"https://${nodeSpec.host}:${nodeSpec.port}"
     } else {
       s"http://${nodeSpec.host}:${nodeSpec.port}"
     }
+
+  // Failsafe: forward only client-recognized options, warn on the rest; copies so options is intact.
+  private def clientOptions(nodeSpec: NodeSpec): util.Map[String, String] = {
+    val (kept, dropped) = nodeSpec.options.asScala.partition { case (key, _) =>
```
Marais: @chernser Can you verify if this is correct. I.e. we have an allow list of options:
`  private val ALLOWED_OPTION_KEYS: Set[String] =
    ClientConfigProperties.values().map(_.getKey).toSet

  private val ALLOWED_OPTION_PREFIXES: Seq[String] =
    Seq(ClientConfigProperties.HTTP_HEADER_PREFIX, ClientConfigProperties.SERVER_SETTING_PREFIX)`

chernser: Yes, this is correct. 

But - there is another prefix `custom_` that can be configured on server. This used to pass custom properties that server do not validate.
Please read more here https://github.com/ClickHouse/clickhouse-java/blob/f95aeb90836c76eb6b331d1722fca9cf1c7abc93/client-v2/src/main/java/com/clickhouse/client/api/ClientConfigProperties.java#L218

and here https://clickhouse.com/docs/concepts/features/configuration/settings/settings-query-level#custom_settings
They have to be processed, too. 

Should we expose logic of configuration validation so connector can call it and get list of unsupported properties instead of throwing exception?

Marais: Thanks. Yes, that will be handy to expose it.

Marais:
> LGTM
@ShimonSte 
I made some changes after Sergei's feedback. In short, we dont allow any connector prps throw, but allow others. Let the java-client warn about. A test will fail once we upgrade to 0.9.7 or up, which protect as against exception. We will then pass the ignore flag to be safe. But now we are on 0.9.5 , so it is not relevant.

ShimonSte (approved):

PR #564, opened 2026-07-16
ci: nightly test against latest ClickHouse Java client (#550)

Marais:
## What / Why

Adds a nightly workflow (`.github/workflows/nightly-java-client.yml`) that runs the connector test suite against the latest ClickHouse Java client **built from source**, to catch breaking changes before an official client release. Implements ClickHouse/spark-clickhouse-connector#550. CI-only — no source or build-config changes.

## How to run

`clickhouse_java_ref` accepts a **branch, tag, or commit SHA** (short SHAs are API-expanded). Runs nightly at 04:00 UTC against `main`, or manually:

```bash
gh workflow run nightly-java-client.yml                                # main
gh workflow run nightly-java-client.yml -f clickhouse_java_ref=v0.9.8  # tag
gh workflow run nightly-java-client.yml -f clickhouse_java_ref=a3ccce4 # commit
```

All four modes were validated on the branch (temp push-trigger + matrix, since removed).

## Findings

The run surfaced a real connector ↔ client incompatibility — the workflow doing its job:

```
ClientMisconfigurationException: Unknown and unmapped config properties: [ssl]
```

`NodeClient` passes the full `nodeSpec.options` map (incl. the connector-level `ssl` key) into the client-v2 `Client.Builder`, which newer clickhouse-java now rejects. Reproduces on released `v0.9.8`. Connector-side fix, out of scope here — follow-up.

## Follow-ups

* Fix `ssl` / connector-only options passed into the client-v2 builder in `NodeClient`.
* Decide on failure notifications (Slack / issue) for the nightlies.

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)


.github/workflows/nightly-java-client.yml
```diff
@@ -0,0 +1,92 @@
...
+      CH_JAVA_REF: ${{ github.event.inputs.clickhouse_java_ref || 'main' }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-java@v4
+        with:
+          distribution: zulu
+          java-version: 17
+          cache: gradle
+
+      - name: Build clickhouse-java from '${{ env.CH_JAVA_REF }}'
```
BentsiLeviav, line 57: IIRC, we publish the nightly Java Client snapshot artifact to some central place - @chernser, could you confirm?

In case we have it ready, this step is redundant.

.github/workflows/nightly-java-client.yml
BentsiLeviav, line 1: @Marais can we make sure all these Java client nightly tests (Spark, Flink, Kafka Connect) are being triggered by the Java Client nightly test?

Marais: @BentsiLeviav I mimicked the kafka nightly test, which builds the connector from source. Agreed, maybe it is better to use the snapshort artifacts produced by the  java-client. I checked if the java-client triggers our connectors, it does not at the moment. We need to think about the scope of this pr if we want to create a chain effect (which is a devops projects). Maybe for the scope of this pr, we take the artifact like you suggested, but trigger on a specific nightly time.  WDYT?

BentsiLeviav: @Marais I'm not sure a "chain effect" is a heavy lift (listen in this workflow to a specific event name and make sure the Java client triggers based on this event). 

The idea of this workflow is to notify the Java Client developers ASAP (AKA - on their nightly build)

@ShimonSte could you ellaborate on what it required from the Glue/Spark chain and from the TDVT/Java client chain?

.github/workflows/nightly-java-client.yml
chernser: yes, we build from main. But last days Sonar changed how we work with nightly - we stop seeing directories for examples and it is not clear what is in the repo. So I do not recommend using nightly for this. 
Head compilation takes not so much time so it is more flexible way.

.github/workflows/nightly-java-client.yml
chernser: @BentsiLeviav 
taking HEAD of java and compiling is the best approach. It should be done on schedule basis. It will significantly reduce WF maintenance work. 
As we put tag on each release and have `VERSION` file this approach can be extended with selective testing.

BentsiLeviav: > taking HEAD of java and compiling is the best approach

Can you share why? 

Generally, +1 for supporting manual testing (dispatch with a specific SHA)

chernser: if nightly: 
- nightly build publishes always with the same version that we cannot match to some commit 
- jar visibility is problematic because we rely only on how maven downloads dependecies
- more things to check 

if HEAD or tag: 
- take at any time 
- we know exact content of the jar because each HEAD has a commit hash 
- easier to track - no need to poll maven for the new version 
- can test different versions by tag or commit hash 
- not dependency on more external service like sonar 
-

BentsiLeviav: >  we rely only on how maven downloads dependecies

I think that the goal would be to have the exact dependency resolution flow customers would have - hence, this doesn't sound like a disadvantage.

> if HEAD or tag:

The two (head/tag vs nightly artifact)  are not supposed to be mutually exclusive. We support them both.

Anyway - @Marais, let's add dispatch support and an option to provide a specific Java client commit/tag; we can keep it as is now and revisit if needed later.

.github/workflows/nightly-java-client.yml
```diff
@@ -0,0 +1,126 @@
...
+        java_ref:
+          - main                                        # branch
+          - v0.9.8                                      # tag
+          - a3ccce456964f288690121f5d4fdfe6dc42a91b5    # full commit SHA
+          - a3ccce4                                     # short commit SHA (API-expanded)
+    env:
+      CLICKHOUSE_IMAGE: clickhouse/clickhouse-server:latest
+      DOCKER_MIN_API_VERSION: "1.24"
+      # TEMP: 'matrix.java_ref ||' prefix — remove with the matrix before merge.
+      CH_JAVA_REF: ${{ matrix.java_ref || github.event.inputs.clickhouse_java_ref || 'main' }}
```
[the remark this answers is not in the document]
Marais: Resolved. The temporary `push` trigger and ref matrix were validation-only and were removed in the cleanup commit (`d558a99`). This comment was against `0426036` (the validation commit); current HEAD runs a single job driven by the `clickhouse_java_ref` dispatch input.

.github/workflows/nightly-java-client.yml
```diff
@@ -0,0 +1,126 @@
...
+          -Dspark_binary_version=4.0
+          -Pclickhouse_jdbc_version=${{ steps.build-java.outputs.version }}
+          -Pclickhouse_client_v2_version=${{ steps.build-java.outputs.version }}
+          -PmavenCentralMirror=https://maven-central.storage-download.googleapis.com/maven2/
+
+      - name: Upload logs
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: log-java-snapshot
```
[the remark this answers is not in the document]
Marais: Resolved. This only applied to the temporary matrix, where parallel jobs shared the `log-java-snapshot` artifact name. The matrix has been removed, so each run is a single job with no name collision.

.github/workflows/nightly-java-client.yml
```diff
@@ -0,0 +1,126 @@
...
+
+      - name: Run tests against the built Java client
+        run: >-
+          ./gradlew clean test --no-daemon --refresh-dependencies
+          --init-script "$RUNNER_TEMP/mavenLocal.init.gradle"
+          -Dscala_binary_version=2.13
+          -Dspark_binary_version=4.0
+          -Pclickhouse_jdbc_version=${{ steps.build-java.outputs.version }}
+          -Pclickhouse_client_v2_version=${{ steps.build-java.outputs.version }}
+          -PmavenCentralMirror=https://maven-central.storage-download.googleapis.com/maven2/
```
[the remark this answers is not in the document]
Marais: Not an issue here. `mavenLocal()` is registered via an init-script `allprojects` hook, which runs *before* `build.gradle`'s `allprojects` repositories, so it ends up **first** in resolution order, not last. Verified against this repo:

```
REPO_ORDER :clickhouse-core:    [MavenLocal, maven, maven2]
REPO_ORDER :clickhouse-core-it: [MavenLocal, maven, maven2]
```

Gradle resolves the requested version from `mavenLocal` first, so the locally-built jars take precedence over Central/snapshots and the build-from-source guarantee holds.

.github/workflows/nightly-java-client.yml
```diff
@@ -0,0 +1,112 @@
...
+            # via the GitHub API, then fetch that.
+            if [[ "$CH_JAVA_REF" =~ ^[0-9a-fA-F]{4,39}$ ]]; then
+              echo "Ref '$CH_JAVA_REF' not directly fetchable; expanding as a short SHA via the GitHub API..."
+              FULL_SHA="$(gh api "repos/ClickHouse/clickhouse-java/commits/$CH_JAVA_REF" --jq '.sha')"
+              echo "Expanded to $FULL_SHA"
+              git fetch --depth 1 origin "$FULL_SHA"
+            else
+              echo "Could not fetch clickhouse-java ref '$CH_JAVA_REF'." >&2
+              exit 1
+            fi
```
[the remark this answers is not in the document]
Marais: Not correct. `git fetch origin <ref>` resolves tags as well as branches (a bare refname is matched against both `refs/heads` and `refs/tags`). Confirmed by the validation run: the `Build clickhouse-java from 'v0.9.8'` job's build step succeeded (only the downstream Gradle step failed, on the unrelated `ssl` incompatibility).

.github/workflows/nightly-java-client.yml
```diff
@@ -0,0 +1,121 @@
...
```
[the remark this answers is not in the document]
Marais: Working as intended. The commit-expansion fallback only runs after a direct fetch of the ref fails. The single misfire case is a *mistyped* branch/tag whose name is coincidentally pure hex (4–39 chars) **and** matches an existing commit prefix — extremely unlikely, and a hex-looking ref is overwhelmingly a commit SHA (the intended case). Any non-hex typo still fails loudly with `exit 1`. Happy to tighten (e.g. require a min length or `--` prefix convention) if preferred, but treating this as negligible.

.github/workflows/nightly-java-client.yml
```diff
@@ -0,0 +1,121 @@
...
+          echo "Building clickhouse-java at commit $(git rev-parse HEAD)"
+          # Installs into the local Maven repository, including the shaded ':all' artifacts the
+          # connector depends on (the shade executions are bound to the default package phase).
+          mvn -B -q -DskipTests -Dmaven.javadoc.skip=true -Dgpg.skip=true clean install
+          VERSION="$(mvn -q -DforceStdout help:evaluate -Dexpression=project.version)"
+          if [[ "$VERSION" == *'${'* || -z "$VERSION" ]]; then
+            VERSION="$(mvn -q -DforceStdout help:evaluate -Dexpression=revision)"
+          fi
+          echo "Built clickhouse-java version: $VERSION"
+          echo "version=$VERSION" >> "$GITHUB_OUTPUT"
```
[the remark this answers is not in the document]
Marais: Fair, low-severity. The primary resolution *is* validated — an empty or `${...}`-containing `project.version` falls back to `revision` — but the fallback value itself isn't re-checked. In practice `revision` resolves for clickhouse-java, so no live failure, but a final non-empty / no-placeholder guard on `$VERSION` before writing `GITHUB_OUTPUT` would be a reasonable hardening. Noting as an optional follow-up.

.github/workflows/nightly-java-client.yml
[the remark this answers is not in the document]
Marais: Addressed in `f5a16ce`: the short-SHA fallback now requires `>=7` hex chars (`^[0-9a-fA-F]{7,40}$`), so a mistyped branch/tag can't be reinterpreted as a commit prefix. Non-hex refs still fail loudly.

.github/workflows/nightly-java-client.yml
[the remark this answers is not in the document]
Marais: Addressed in `f5a16ce`: after the `project.version` -> `revision` fallback, `$VERSION` is now re-checked for empty / unresolved `${...}` and the step exits 1 with a clear message before writing `GITHUB_OUTPUT`.

.github/workflows/nightly-java-client.yml
BentsiLeviav: @chernser, in addition, should we have this test triggered by any Java repo workflow? The minimum I had in mind is during release testing (no need to block the release, but at least include it so the relevant Java engineer will be aware).

Marais: @BentsiLeviav I made it support these modes:
`gh workflow run nightly-java-client.yml                                # main
gh workflow run nightly-java-client.yml -f clickhouse_java_ref=v0.9.8  # tag
gh workflow run nightly-java-client.yml -f clickhouse_java_ref=a3ccce4 # commit`

Note: Release 0.9.8 is currently not compatible with us, until my SSL fix pr goes in.

chernser: @BentsiLeviav
This workflow should be only be run by schedule. 
We have no specific WF before release. Besides we may do many merges to main so process may be not so easy to handle (you need to take latest).

BentsiLeviav: @chernser, so what should trigger this in the Java client repo?

The problem with "nightly only" is that when you merge some changes and release them the same day, they are not covered.

BentsiLeviav (approved):
Let's first understand why the Claude tests are failing. If this is related to data sync between replicas - we have settings to help with that.

Marais:
> Let's first understand why the Claude tests are failing. If this is related to data sync between replicas - we have settings to help with that.

not sure what you mean by "Claude tests". The failures are the arrow tests - not related. The sync between replicas, I think we should track this seperately since tis happens on main.

Marais:
> > Let's first understand why the Claude tests are failing. If this is related to data sync between replicas - we have settings to help with that.
> 
> not sure what you mean by "Claude tests". The failures are the arrow tests - not related. The sync between replicas, I think we should track this seperately since tis happens on main.

I will merge, but noted the left overs. e.g. trigger from jdbc connecor. resolve flakyness.

.github/workflows/nightly-java-client.yml
chernser: As we talked- it will be nightly or release. Later we will add some integration point to trigger such tests.

PR #579, opened 2026-08-26
fix(read): keep runtime statistics out of table identity

ShimonSte:
`ClickHouseTable` is a case class, so its generated equality includes `spec.total_rows`, `spec.total_bytes` and the `lifetime_*` counters — runtime statistics that move as data commits.

Spark finds cached data by comparing plans, and a `DataSourceV2Relation` carries this `Table`, so an identity that shifts with the data silently defeats plan reuse: `CACHE TABLE`, reused exchanges, self-join detection. On Cloud the `cache table` integration test failed intermittently with `cachedPlan.isDefined was false`.

This compares on the stable fields instead. The statistics stay in `properties()`, where they're informational; only equality ignores them. `functionCatalogUsable` stays in the key — it genuinely separates catalog tables from `format("clickhouse")` ones — while `functionRegistry` is excluded as one instance per catalog, already implied by `node`/`cluster`.

**Caveat worth reviewing:** keeping runtime statistics out of a table's identity is correct on its own terms, but it is **not proven** to be the cause of that flake. I couldn't reproduce it outside CI, and I disproved my own first hypothesis — every catalog load I sampled from `query_log` was served by a single replica, ruling out cross-replica divergence as the mechanism there.

Adds one pure unit test. Verified on all four Spark versions: 64/88/88/95 unit tests, 0 failures.


spark-4.0/clickhouse-spark/src/main/scala/com/clickhouse/spark/ClickHouseTable.scala
```diff
@@ -73,6 +73,33 @@ case class ClickHouseTable(
...
 
+  /**
+   * Runtime statistics are not part of a table's identity. `total_rows`, `total_bytes` and the
+   * `lifetime_*` counters move as data commits, so letting them into equality makes two catalog
+   * loads of the same table unequal. Spark compares plans to find cached data, and a
+   * DataSourceV2Relation carries this Table, so a shifting identity silently defeats plan reuse
+   * (CACHE TABLE, reused exchanges, self-join detection). They stay in `properties()`; only
+   * equality ignores them.
+   */
+  private def identityKey =
```
Marais, line 86: 1. Distributed tables aren't fixed. identityKey includes cluster, and ClusterSpec/ShardSpec hold Arrays, which Scala compares by reference. Two loads via different catalog instances (format("clickhouse") builds one per getTable; the CacheManager is app-wide) never compare equal, so the cache miss survives for exactly the Distributed case. Needs value equality on those specs, plus a test with cluster = Some(...) — the new test only covers cluster = None.
2. Residual flake risk: metadata_modification_time, data_paths, metadata_path are replica-local and stay in the key, so on multi-replica Cloud two loads can still differ. Given the root cause is unproven, consider an allow-list of stable fields instead of denying four.

ShimonSte: Both correct — thanks, the first one was a real hole and my test gave false confidence.

**1.** Confirmed: `ClusterSpec.shards` and `ShardSpec.replicas` are Arrays with no `equals` override, so case class equality compared them by reference. And `ClickHouseTableProvider.getTable` calls `createCatalog(options)` per call, so two loads via `format("clickhouse")` really do hold different instances. The catalog path happened to work only because Spark caches catalog instances and `clusterSpecs` is built once in `initialize` — same reference — which is why the flake looked fixed. Gave both specs value equality in 6016cf2 (`compare` left keyed on `num`, since it orders shards within one cluster). `ReplicaSpec`, `TableEngineSpec` and `NodeSpec` have no arrays, so those two were the only ones.

**2.** Switched to an allow-list in a5f64c3: node, cluster, tz, database, name, uuid, engine, engineSpec, functionCatalogUsable — so those three fields are simply out of the key. On the premise itself I couldn't reproduce divergence (steady state: identical across all three replicas, `metadata_path` empty on Cloud; 0/12 divergence immediately after `ALTER TABLE ADD COLUMN`), though my probes on this cluster have shown weak power before, so I wouldn't read that as absence. The design argument stands on its own regardless.

Tests are now four, including `cluster = Some(...)` with a `DistributedEngineSpec` and a negative case that different tables still compare unequal. I verified each has teeth by reverting its fix: the cluster test fails without the value equality, the replica-local test fails with the subtractive key.

clickhouse-core/src/main/scala/com/clickhouse/spark/spec/NodeSpec.scala
```diff
@@ -98,6 +106,14 @@ case class ClusterSpec(
 
   override def toString: String = s"cluster: $name, shards: [${shards.mkString(", ")}]"
 
+  // Arrays compare by reference, so without this two specs describing the same cluster differ.
+  override def equals(other: Any): Boolean = other match {
+    case that: ClusterSpec => name == that.name && shards.toSeq == that.shards.toSeq
+    case _ => false
+  }
+
+  override def hashCode(): Int = (name, shards.toSeq).hashCode()
```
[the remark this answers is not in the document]
ShimonSte: Agreed and fixed in 81a14cd. Confirmed the premise: `queryClusterSpecs` has no `ORDER BY` and builds both arrays via `groupBy(...).map(...).toArray`, so their order follows an unordered result.

Worth adding why it is reachable rather than theoretical: for four or fewer entries Scala's `Map1`–`Map4` iterate in insertion order, so the array order tracks the row order ClickHouse happens to return. Our own CI service is 1 shard with 3 replicas — exactly that shape — and it was rescaled mid-investigation, which is the kind of event that reshuffles those rows.

Now compares the sorted arrays (`ClusterSpec` reuses its precomputed `sortedShards`; both `ShardSpec` and `ReplicaSpec` are already `Ordered` by `num`).

One note in case it helps future reviews: my first attempt at the test reordered hosts with a helper that assigned replica numbers by position, so it built a genuinely different topology (replica 1 pointing at another host) and correctly stayed unequal — the test failed for the right reason, not because sorting was insufficient. The committed version reorders while keeping each `num` paired with its host, and also asserts that same-hosts-with-swapped-numbers still differs, so the fix cannot pass by over-normalising.

spark-4.0/clickhouse-spark/src/main/scala/com/clickhouse/spark/ClickHouseTable.scala
ShimonSte: Changed it, added to cluster and shard spec + using an allow list

clickhouse-core/src/main/scala/com/clickhouse/spark/spec/NodeSpec.scala
```diff
@@ -88,6 +88,15 @@ case class ShardSpec(
 
   override def toString: String = s"shard_num: $num, shard_weight: $weight, replicas: [${replicas.mkString(", ")}]"
 
+  // Arrays compare by reference; see ClusterSpec#equals. `compare` stays keyed on `num` alone.
+  override def equals(other: Any): Boolean = other match {
+    case that: ShardSpec =>
+      num == that.num && weight == that.weight && replicas.sorted.toSeq == that.replicas.sorted.toSeq
```
Marais: Nit: ClusterSpec reuses its precomputed sortedShards, but here equals/hashCode re-sort replicas on every call — and since a cluster comparison walks every shard, each table-identity check sorts every shard's replicas on both sides. A private lazy val sortedReplicas mirroring the sortedShards pattern would make it allocation-free after first use. Fine as a follow-up; replica arrays are tiny.

Marais (approved):

clickhouse-core/src/main/scala/com/clickhouse/spark/spec/NodeSpec.scala
ShimonSte: Done in b319ffc0, with one deviation worth flagging: both specs now hold the sorted collection as a `Seq` rather than an `Array`.

Mirroring `sortedShards` exactly would not have reached allocation-free — `Array#toSeq` allocates a wrapper on every call, so `ClusterSpec.equals` was already paying two of those per comparison even with the array precomputed. Storing the `Seq` removes that too. `sortedShards` stays as-is, since `shardWeightUpperBounds` and `nodes` need the array form.

Also folded in the two `nodes` lazy vals, which were re-sorting rather than reusing the precomputed values (`shards.sorted.flatMap(...)` one line above `sortedShards`, and the same in `ShardSpec`). Once per instance rather than per call, so not a hot path, but they were the same inconsistency.

After first use a comparison now sorts and allocates nothing, where an S-shard cluster previously did 2S sorts plus wrappers.

On the framing: I agree it is worth doing, though for consistency rather than speed — this runs at plan-comparison time, not per row, unlike `shardWeightUpperBounds` whose comment cites the per-row write path. Two siblings in one file solving the same problem two ways is the part that would cost a future reader time.

Verified the order test still has teeth after the refactor by removing the sort from `sortedReplicas` and confirming it fails, so the assertion has not gone vacuous.

Marais (approved):

PR #580, opened 2026-08-26
fix(read): cover partitions missing from a stale partition listing

ShimonSte:
A scan plans one task per partition from `queryPartitionSpec`, each filtering `_partition_id = '<id>'`, so a scan returns the union of those predicates - correctness depends on the listing being complete. It reads `system.parts`, a per-replica view with no consistency control.

Observed in CI: two listings of the same table one second apart, on the same replica, returned **3 and 4 partitions**. The 3-partition plan issued three tasks; A partition was never queried, so the read returned 3 rows instead of 4 - silently, no error. Three tests failed this way. `select_sequential_consistency` can't help: the query for the missing partition is never issued, because the planner never learned it existed. In edge cases this affects users too, reading a partitioned table on Cloud can return silently incomplete results.

Fix: plan one extra task per listing, filtering `_partition_id NOT IN (<listed ids>)`, so the predicates cover the table by construction:

```
(A ∪ B ∪ C) ∪ ¬(A ∪ B ∪ C) = everything
```

An incomplete listing then costs read parallelism instead of rows, for one usually-empty task per scan.

Two guards worth review: the complement is skipped when a listing already reads the whole table (`NoPartitionSpec` yields `1=1`), which would otherwise read every row twice, and when not splitting by `_partition_id`, since that mode predicates on `partition_key = value` where multi-column keys become tuples and `NOT IN` NULL semantics are ill-defined. The single-partition warning now counts only listed partitions.

A complement that matches nothing reads zero rows - except under a pushed-down aggregation with no `GROUP BY`, which still returns one row holding the empty-set aggregate (`min` of no rows is 0 when its on a non nullable path) that Spark folds into the final value. Hence `HAVING count() > 0` there; an earlier revision without it broke `push down aggregation` on every matrix job. The `GROUP BY` branch is verified against ClickHouse directly, not through the connector - no test here exercises a pushed-down `GROUP BY`, since the suite's grouped cases use `DISTINCT` aggregates that `compileAggregate` cannot compile.

Considered `clusterAllReplicas` for the listing: it lowers the probability of a short listing rather than removing it, and one unreachable replica then fails every read plan - while `skip_unavailable_shards=1` restores availability by silently dropping a replica's contribution, reintroducing the bug. for now i decided not to solve the issue with `clusterAllReplicas`, open for discussion here.


spark-3.3/clickhouse-spark/src/main/scala/com/clickhouse/spark/read/ClickHouseRead.scala
```diff
@@ -196,16 +196,18 @@ class ClickHouseBatchScan(scanJob: ScanJobDescription) extends Scan with Batch
       case DistributedEngineSpec(_, _, local_db, local_table, _, _) if scanJob.readOptions.convertDistributedToLocal =>
         scanJob.cluster.get.shards.flatMap { shardSpec =>
           Utils.tryWithResource(NodeClient(shardSpec.nodes.head, queryTimeoutMs)) { implicit nodeClient: NodeClient =>
-            queryPartitionSpec(local_db, local_table).map { partitionSpec =>
+            // complemented per shard: each shard is listed separately, so one complement over the
+            // flattened result would exclude the other shards' ids and be pinned to a single shard
+            withComplement(queryPartitionSpec(local_db, local_table).map { partitionSpec =>
```
Marais, line 201: The `NOT IN` list makes the query text grow with partition count: past ~7.3k hash-style ids the complement query exceeds the default `max_query_size` (256 KB) and every read of the table fails — repro'd on 24.8 (7200 ids parse, 7300 → `Code: 62`). Inline `SETTINGS` can't raise it (parse-time limit).

Suggest a cap: over a threshold, fall back to a single whole-table task (`NoPartitionSpec` path) — still correct, just less parallelism. Applies to all four profiles.

spark-3.5/clickhouse-spark/src/main/scala/com/clickhouse/spark/read/ClickHouseReader.scala
```diff
@@ -38,6 +38,15 @@ abstract class ClickHouseReader[Record](
...
   private val readSettings: Option[String] = conf.getConf(READ_SETTINGS)
 
+  /**
+   * A complement partition matches nothing whenever the partition listing was complete. That reads
+   * as zero rows normally and for a GROUP BY aggregation, but a pushed-down aggregation without
+   * GROUP BY still returns one row holding the aggregate of the empty set (`min` of no rows is 0),
+   * which would corrupt the value Spark combines across partitions. Drop that row.
+   */
+  private val emptyAggregateGuard: String =
+    if (part.complementOf.isDefined && scanJob.groupByClause.isDefined) "HAVING count() > 0" else ""
```
Marais, line 48: **[critical severity, but pre-existing — follow-up issue is fine]** The guard only fires for the complement, but listed partitions hit the same empty-set row: `MIN(id) ... WHERE _partition_id='p' AND <filter>` with no matches returns 0 and corrupts the global MIN/MAX. Note extending the guard naively flips pushed COUNT on an empty table from 0 to NULL (Spark sums partials).

spark-3.5/clickhouse-spark/src/main/scala/com/clickhouse/spark/read/ClickHouseRead.scala
```diff
@@ -222,22 +224,45 @@ class ClickHouseBatchScan(scanJob: ScanJobDescription) extends Scan with Batch
...
       log.warn(s"Reading $database.$table as a single partition, which may impact read performance")
     }
     partitions
   }
 
+  /**
+   * Appends one input partition covering every partition absent from `parts`. The partition listing
+   * comes from `system.parts`, which is only eventually consistent on ClickHouse Cloud: a listing
+   * that misses a partition would otherwise prune it away and silently return incomplete results.
+   * The complement makes the task predicates exhaustive, so a stale listing costs read parallelism
```
Marais, line 248: **[should fix before merge]** "Exhaustive" only holds against one snapshot: each reader task picks its own server, so a lagging replica serving a *listed* partition can still return 0 rows (what #578 measured on Cloud). Worth softening here and in the changelog; `select_sequential_consistency` as a follow-up?

spark-3.5/clickhouse-spark-it/src/test/scala/org/apache/spark/sql/clickhouse/cluster/ClickHouseClusterReadSuite.scala
```diff
@@ -47,6 +48,32 @@ class ClickHouseClusterReadSuite extends SparkClickHouseClusterTest {
     }
   }
 
+  test("each shard is complemented from its own partition listing") {
+    // a single complement over the flattened result would carry every shard's ids and sit on one
+    // shard, so the rest of the cluster would keep the exposure the complement exists to remove
+    withSimpleDistTable("single_replica", "db_compl", "t_dist", true) { (_, db, tbl_dist, _) =>
+      withSQLConf(READ_DISTRIBUTED_CONVERT_LOCAL.key -> "true") {
+        val df = spark.sql(s"SELECT id FROM $db.$tbl_dist")
+        df.collect()
```
Marais, line 57: **[should fix before merge]** Nothing asserts the data — a complement that wrongly matches listed partitions would duplicate every row and still pass. `checkAnswer(df, Seq(Row(1), Row(2), Row(3), Row(4)))`?

spark-3.5/clickhouse-spark/src/test/scala/com/clickhouse/spark/read/ClickHouseBatchScanSuite.scala
```diff
@@ -46,6 +48,38 @@ class ClickHouseBatchScanSuite extends AnyFunSuite with Log4j2CaptureHelper {
...
+    val complement = inputPartition(NoPartitionSpec, complementOf = Some(Seq("20220411", "20220412")))
+    assert(complement.partFilterExpr === "_partition_id NOT IN ('20220411', '20220412')")
+  }
+
+  test("a listed input partition still filters by its own partition id") {
+    val listed = inputPartition(PartitionSpec("2022-04-11", "20220411", 1, 1))
+    assert(listed.partFilterExpr === "_partition_id = '20220411'")
+  }
+
+  test("no complement is planned when the listing collapsed to a whole-table scan") {
```
Marais, line 61: **[should fix before merge]** This never reaches `withComplement`: `scanJob()` leaves `convertDistributedToLocal` off, so the bare Distributed branch builds `NoPartitionSpec` directly — deleting `|| readsWholeTable` keeps the suite green. Test `withComplement` directly instead?

spark-3.5/clickhouse-spark/src/main/scala/com/clickhouse/spark/read/ClickHouseRead.scala
```diff
@@ -222,22 +224,45 @@ class ClickHouseBatchScan(scanJob: ScanJobDescription) extends Scan with Batch
...
+   * instead of rows.
+   *
+   * Skipped unless splitting by `_partition_id` (the partition-value predicates have known bugs and
+   * ill-defined NOT IN semantics), and skipped when the listing collapsed to [[NoPartitionSpec]],
+   * which already scans the whole table.
+   */
+  private def withComplement(parts: Seq[ClickHouseInputPartition]): Seq[ClickHouseInputPartition] = {
```
Marais, line 255: **[non-blocking]** On by default with no opt-out: +1 task/connection/no-op query per shard per scan, and on 3.3/3.4 a one-partition table now plans 2, losing the `SinglePartition` shortcut (extra exchange under global aggregates). Worth a conf to disable?

spark-3.5/clickhouse-spark-it/src/test/scala/org/apache/spark/sql/clickhouse/cluster/ClickHouseClusterReadSuite.scala
```diff
@@ -47,6 +48,32 @@ class ClickHouseClusterReadSuite extends SparkClickHouseClusterTest {
...
+        val scan = df.queryExecution.executedPlan.collectFirst {
```
Marais, line 58: **[non-blocking]** `.collectFirst { ... }.get` dies with `None.get` under AQE — use `queryExecution.sparkPlan` like other tests in this suite. And the count assertion goes off-by-one if a shard's listing collapses to `NoPartitionSpec` (reachable: the fixture's async Distributed write is guarded only by `sleep(2000)`).

spark-3.5/clickhouse-spark/src/main/scala/com/clickhouse/spark/read/ClickHouseRead.scala
```diff
@@ -222,22 +224,45 @@ class ClickHouseBatchScan(scanJob: ScanJobDescription) extends Scan with Batch
...
+    // NoPartitionSpec already yields a `1=1` predicate covering the whole table, so complementing it
+    // would plan a second whole-table scan and read every row twice
+    val readsWholeTable = parts.exists(_.partition == NoPartitionSpec)
+    if (!scanJob.readOptions.splitByPartitionId || parts.isEmpty || readsWholeTable) {
```
Marais, line 259: **[non-blocking]** `parts.isEmpty` is dead (`queryPartitionSpec` collapses empty listings to `NoPartitionSpec`) and fails open — zero tasks for the shard = silent row loss if that ever changes. Drop it, or fall back to a whole-table partition.

spark-3.5/clickhouse-spark/src/main/scala/com/clickhouse/spark/read/ClickHouseReader.scala
```diff
@@ -59,6 +68,7 @@ abstract class ClickHouseReader[Record](
        |FROM `$database`.`$table`
        |WHERE (${part.partFilterExpr}) AND (${scanJob.filtersExpr})
        |${scanJob.groupByClause.getOrElse("")}
+       |$emptyAggregateGuard
```
Marais, line 71: **[non-blocking]** No test pins this SQL — a unit test asserting the generated query for a complement + pushed aggregate would cover the guard condition and the `HAVING` placement.

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
 ### Fixed
+- Reads no longer silently drop rows when the partition listing is incomplete ([#580](https://github.com/ClickHouse/spark-clickhouse-connector/pull/580)). A scan plans one input partition per partition returned by `system.parts`, each filtering on `_partition_id`, so the rows a scan returns are the union of those filters and a listing that misses a partition prunes it away with no error. `system.parts` is only eventually consistent on ClickHouse Cloud, where each operation opens its own connection and may be served by a replica that has not caught up: two listings of the same table one second apart were observed returning 3 and 4 partitions, and the read built from the shorter one returned 3 of 4 rows. Scans now plan one additional input partition filtering `_partition_id NOT IN (<listed ids>)`, so the filters cover the whole table by construction and an incomplete listing costs read parallelism instead of rows. It is skipped when a scan already reads the whole table and when `spark.clickhouse.read.splitByPartitionId=false`. Applies to all Spark profiles (3.3/3.4/3.5/4.0).
```
Marais, line 33: **[nit]** Under `convertDistributedToLocal` it's one complement per shard, not one per scan (the IT test asserts `complements.length > 1`) — "one per partition listing (per shard)".

spark-3.5/clickhouse-spark/src/main/scala/com/clickhouse/spark/read/ClickHouseRead.scala
```diff
@@ -198,16 +198,18 @@ class ClickHouseBatchScan(scanJob: ScanJobDescription) extends Scan with Batch
...
+            // flattened result would exclude the other shards' ids and be pinned to a single shard
+            withComplement(queryPartitionSpec(local_db, local_table).map { partitionSpec =>
               ClickHouseInputPartition(
                 scanJob.localTableSpec.get,
                 partitionSpec,
                 scanJob.readOptions.splitByPartitionId,
                 shardSpec // TODO pickup preferred
               )
-            }
+            }.toSeq)
```
Marais, line 210: **[nit]** The added `.toSeq` (×2) and the `.toArray` on the flatMap are no-ops — compiles clean without them on 2.12 and 2.13.

spark-3.5/clickhouse-spark/src/main/scala/com/clickhouse/spark/read/ClickHouseRead.scala
```diff
@@ -222,22 +224,45 @@ class ClickHouseBatchScan(scanJob: ScanJobDescription) extends Scan with Batch
...
     }
-    if (partitions.length == 1) {
+    // counted on the listed partitions only: a complement is not a partition of the table
+    if (partitions.count(_.complementOf.isEmpty) == 1) {
       log.warn(s"Reading $database.$table as a single partition, which may impact read performance")
     }
     partitions
   }
 
+  /**
```
Marais, line 244: **[nit]** Could these scaladocs/block comments be one-liners? The rationale already lives in the PR/changelog, and it's ×4 trees to keep in sync.

PR #581, opened 2026-08-31
fix(read): union partition listings across a table's replicas

ShimonSte:
A scan plans one input partition per partition reported by `queryPartitionSpec`, each filtering on `_partition_id`, so a listing that omits a partition prunes it out of the scan and silently drops its rows.

`system.parts` reflects only the server answering the query, and the connector opens a fresh `NodeClient` per operation — so on Cloud each listing may reach a different, independently-lagging replica. Observed in CI: two listings of the same table one second apart returned **3 and 4 partitions**, and the read built from the shorter one returned 3 of 4 rows, silently and with no error. **This affects users, not just tests.**

Fix: read whole-table listings from `clusterAllReplicas(<cluster>, system.parts)`, deduplicated by part name. Without the dedupe a part reported by every replica is counted once per replica — measured **6 rows where the table held 2**.

### Choosing the cluster

Discovered, not configured: `system.clusters` gives the clusters containing this server (`is_local`), and the one with the fewest shards is picked so a read does not fan out to shards that cannot hold the table's data. Single-server clusters are ignored, so a non-replicated deployment pays nothing.

`is_local` is the safety property: `clusterAllReplicas` always executes on the local member, so the union is a **superset** of that server's own view — the listing the connector read before. It can only add a partition the local view was missing, never drop one. A configurable cluster name was tried and removed: it carries no such guarantee, and a cluster excluding this server could make the union *smaller*, which is the one way this change could lose rows.

`remote()` over the shard's own replicas would be more precise, but the initiator does not forward the client's credentials (`Code: 516`).

### Settings and failure handling

The union runs with `skip_unavailable_shards=1, max_execution_time=10, timeout_overflow_mode='throw'`. Where those settings are permitted, skipping keeps the union answering while a replica is down instead of failing with `Code: 279` on every scan. The overflow mode is pinned because `max_execution_time` only *throws* under the default mode; under `break` a query that hits its limit returns a truncated result with **HTTP 200 and no error**, which would be read as a complete listing — measured on staging, a 1000-group aggregate returned **21 groups**. A `readonly=1` user may not set the first two (`Code: 164`), so discovery probes them once and retries with the pin alone.

Any other failure falls back to the answering server's own view, which is what the connector read before, so an unfitting topology degrades rather than failing.

### Where it does not apply

Only scans filtering by `_partition_id`. With `read.splitByPartitionId` off, tasks filter on a partition *value* taken from `any(partition)`, rendered in the answering server's timezone — a replica elsewhere renders a `DateTime` partition differently. Per-shard listings under `read.distributed.convertLocal` are left alone too, since the union would report every shard's partitions to each shard; so is partition-identifier listing (`SHOW PARTITIONS`), which consumes only that value. Both this and the kill switch `read.partitionListing.unionReplicas` are honoured as session conf and as per-read `.option(...)`.

### Empty-set aggregates

ClickHouse returns a non-nullable column's default rather than NULL for an aggregate over no rows, so a task whose filters matched nothing contributed a `0` that Spark folded into the final value. The union made this reachable — it can report partitions this server does not hold — but it is reachable without the union too, so this also fixes a pre-existing bug: `SELECT MIN(id) FROM t WHERE id > 999` returns `0` rather than NULL on `main`.

Fixed by pushing `MIN`, `MAX` and `SUM` as `minOrNull`, `maxOrNull` and `sumOrNull`. `COUNT` deliberately keeps its plain form: an earlier revision dropped the empty row with `HAVING count() > 0`, which corrupted COUNT — Spark merges partials with a `Sum`, `Sum` of zero rows is NULL, and COUNT's output attribute is non-nullable, so `COUNT(*)` read as `0` through `collect()` yet **vanished** under `filter("c = 0")`. The tests had missed it because `checkAnswer` uses `collect()`, the one consumer that masks it; they now assert through a filter too.

### Scope, stated plainly

This narrows the window rather than closing it. A task still reads from whichever server its own connection reaches, so a lagging replica can return no rows for a partition that *was* listed — read consistency rather than planning. Accepting that is deliberate: it is the same-instant race, plus the case where the replica holding a write is unreachable.

### Relationship to #580

#580 attacks the same bug by planning an extra input partition filtering `_partition_id NOT IN (<listed ids>)`, so the plan covers the table regardless of the listing. Stronger, but it added failure modes: past ~7.3k partitions the predicate exceeds `max_query_size` and every read fails, and the complement had to be built per shard. This fixes the listing itself rather than a consequence downstream of it. The two are alternatives; whichever is preferred, the other should be closed.

### Testing

`withReplicatedTable` creates a `ReplicatedMergeTree` on each shard's two replicas, which the suite previously had no fixture for. That makes both halves testable in CI:

- **Deduplication.** After `SYSTEM SYNC REPLICA` both replicas report both parts; removing the `GROUP BY name` fails the test with `4 did not equal 2`.
- **The recovery itself.** `SYSTEM STOP FETCHES` stalls the connected server and the next write goes to its peer, so the partition is genuinely absent from this server's `system.parts` and present in the union. Disabling the union fails the test.

Every fix was verified by reverting it and watching the test fail, not just by passing. Assertions read the planned input partitions rather than collected rows, because a scan of the stalled server still returns no rows for that partition — the read-consistency limit noted above.

Also covered: both forms of both switches; that a cluster spanning a shard we do not hold returns identical rows to reading with the union off; that a table present only on the connected server reads correctly, its peers contributing nothing; and that a pushed-down aggregate is not corrupted by partitions this server does not hold.

Not covered in CI: the fallback path, which needs an unreachable replica, and the overflow-mode pin, which needs a deterministic timeout. Both were verified against a live 3-replica service instead.

Verified on every supported combination (3.3/3.4/3.5 on Scala 2.12 and 2.13, 4.0 on 2.13): unit, integration and `clickhouse-core` suites green. The ClickHouse Cloud workflow was green on the last code commit.


spark-3.5/clickhouse-spark/src/main/scala/com/clickhouse/spark/SQLHelper.scala
```diff
@@ -73,16 +73,22 @@ trait SQLHelper {
...
+   * SQL's NULL instead. `COUNT` keeps its plain form: 0 is its correct empty-set value, and Spark
+   * merges it with a `Sum` whose output attribute is non-nullable, so a NULL there corrupts it.
+   */
   def compileAggregate(aggFunction: AggregateFunc): Option[String] =
     aggFunction match {
       case min: Min if min.column.isInstanceOf[NamedReference] =>
         val col = min.column.asInstanceOf[NamedReference]
         if (col.fieldNames().length != 1) return None
-        Some(s"MIN(${quoted(col.fieldNames.head)})")
+        Some(s"minOrNull(${quoted(col.fieldNames.head)})")
```
Marais: `minOrNull`/`maxOrNull` return `Nullable(T)`, and ClickHouse can't put Array/Tuple/Map inside Nullable:

`SELECT minOrNull(tags) FROM t WHERE 1=0` → `Code: 43. Nested type Array(String) cannot be inside Nullable type`

On main `MIN(tags)` over an `Array(String)` column pushed down fine. With this change the probe in `pushAggregation` fails, so every such query logs `ERROR Failed to push down aggregation to ClickHouse` with a stack trace and the whole column is read into Spark for it to aggregate. Same for `maxOrNull` on L91.

Can we keep plain `MIN`/`MAX` for Array/Struct/Map columns? `compileAggregate` only sees the column name, so the gate probably has to live in `pushAggregation` (which has `physicalSchema`), or the schema gets passed in. Applies to all four trees.

spark-3.5/clickhouse-spark/src/main/scala/com/clickhouse/spark/ClickHouseHelper.scala
```diff
@@ -332,6 +331,130 @@ trait ClickHouseHelper extends SQLConfHelper with Logging {
...
+          s"SETTINGS $settings"
+        )
+        Try {
+          val output = nodeClient.syncQueryAndCheckOutputJSONEachRow(union)
+          // Scala 2.12 parses the rows lazily, so a mid-stream failure must surface inside the Try
+          output.rows
+          output
+        } match {
+          case Success(output) => output
+          case Failure(cause) =>
```
Marais, line 385: `clusterAllReplicas` needs the `REMOTE` grant. A user with only `SELECT` grants fails the union here with `Code: 497 ... it's necessary to have the grant REMOTE ON *.*` on every scan (verified on 25.3; the plain `system.parts` listing on main and the data read itself work without it).

Cloud's built-in roles (`default_role`, `sql_console_read_only`) include `READ, WRITE ON REMOTE`, so console-created users are fine. The exposure is hand-granted users, the normal shape for service accounts on self-managed clusters, and `REMOTE` is a grant security teams withhold deliberately since it opens `remote()` to any host.

For that user each read now does discovery + the rejected union + the own view, logs this WARN with a full stack trace every time, and never gets the fix. Nothing remembers the failure across reads because `discoveredListingCluster` lives on the per-query `ClickHouseBatchScan` instance.

Can we stop unioning after a permission failure (check `code` for 497/516 and remember it per node, e.g. keyed by `NodeSpec`, or resolve the listing cluster once in the catalog and thread it through `ScanJobDescription`) and log once, without the stack trace?

